PR #19, opened 2014-12-10
Initial API mock

srri-zz:
Initial implementation of django rest framework


dispatch/api/serializers.py
```diff
@@ -0,0 +1,20 @@
...
+from rest_framework import serializers
+
+#class ArticleSerializer(serializers.HyperlinkedModelSerializer):
+#   class Meta:
+#        model = Article
+#        fields = ('headline')
+
+class UserSerializer(serializers.HyperlinkedModelSerializer):
+    class Meta:
+        User = get_user_model()
```
psiemens: I don't think User should be inside Meta

srri-zz: Well I disagree :-1:

srri-zz: But i'll move it

psiemens: :imp:

psiemens:
:+1:

Issue #10, opened 2014-12-10
Sub task: Creating model for image(s)


srri-zz:
Requirements:
- Create a model for a gallery or image(s)

Issue #7, opened 2014-12-10
Sub task: Creating model for article 


srri-zz:
Requirements:
- Create a model (with django in mind) for what an article contains and means

PR #21, opened 2014-12-10
content models

blx:
Adds basic models for Article and Image. (Progress on #10 and #7 but not closing them yet since they'll need more work.)

Will need to come back to this once `user-management` gets merged to uncomment the `User` dependencies, but making the PR now so the models at least exist in `develop`.


srri-zz:
Looks good! +1

srri-zz:
Merge when you want @blx

Issue #14, opened 2014-12-10
Sub task: Add users


srri-zz:
Requirements:
- Ability to add users to a database

Issue #13, opened 2014-12-10
User management


srri-zz:
Requirements:
- Be able to manage users, create users, delete users, modify users.

psiemens:
Completed to all the standards laid out in v 0.1 -- marking as done for now.

Issue #9, opened 2014-12-10
Sub task: Creating model for user


srri-zz:
Requirements:
- create model for a basic user

PR #22, opened 2014-12-10
User management + auth

psiemens:
Adds basic user management, as well as support for contributors and contributor types. Progress on #13, #9, and #14.

### DIFF
--- a/dispatch/core/models.py
+++ b/dispatch/core/models.py
@@ -1,0 +1,20 @@
+from django.db import models
+from django.contrib.auth.models import AbstractBaseUser, BaseUserManager
+
+class UserManager(BaseUserManager):
+    def create_user(self, email, password=None):
+        if not email:
+            raise ValueError('User must have a valid username')
+
+        user = User(email=email)
+        user.set_password(password)
+
+        return user
+
+class User(AbstractBaseUser):
+    email = models.CharField(max_length=255, unique=True)
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    USERNAME_FIELD = 'email'
+    objects = UserManager()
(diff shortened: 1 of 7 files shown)


psiemens:
Actually, you could probably say that all of those tasks are complete. You can add/delete/modify users through the default Django admin. Not sure if this counts, but it works for now.

psiemens:
To better illustrate the relationships I formed: 

user_diagram

A User can be connected to a Person, but a Person doesn't have to be a User. A Person can be a Contributor, and a Contributor can be any number of ContributorTypes (photographer, author, illustrator, etc).

srri-zz:
+10 beautiful diagram, try and resolve your merge conflicts locally though

dispatch/core/models.py
```diff
@@ -0,0 +1,84 @@
...
+            raise ValueError('User must have a valid email address')
+
+        if not self.is_valid_password(password):
+            raise ValueError('Password is invalid')
+
+        user = User(email=email,is_admin=is_admin, is_active=is_active, is_superuser=is_superuser)
+        user.set_password(password)
+        user.save()
+
+        return user
```
srri-zz, line 17: Should it return user or should it just return a boolean value for success/fail in saving the user to the db?

psiemens:
@Captainkrtek Cool, I'll try to figure that out

srri-zz:
Otherwise looks good, see my one comment. I can also try to do the merge if you have issues. :+1:

dispatch/core/models.py
psiemens, line 17: It's supposed to return the new user instance that's created. For example, this allows you to create a new user and then associate them with other models (i.e. make them a contributor). I think it's also what Django expects as a return type as this is an override of a default function.

I understand where you're coming from, though. This method could be improved by adding a try/catch around the user.save() method, which I believe throws an exception if the save is unsuccessful.

PR #23, opened 2014-12-10
Big cleanup

psiemens:
I moved everything up a directory and put all of the apps into a package. Let me know what you think


srri-zz:
Did you test that the API still works since it has no tests yet? Other than that, +1 :+1:

apps/api/serializers.py
```diff
@@ -11,7 +11,7 @@
 class UserSerializer(serializers.HyperlinkedModelSerializer):
     class Meta:
         model = get_user_model()
-        fields = ('url', 'username', 'email', 'groups')
```
psiemens, line 14: Taking out username because we're using email as the username field. Groups hasn't been implemented in the user model yet, so I had to take it out. We can put it back in later.

psiemens:
Whoops, good call. See my line note above. Tested and working.

apps/api/serializers.py
srri-zz, line 14: For the most part these were just to get a model working with the API, we'll likely nuke them and add our own models or modify them. Just wanted to test out the rest_framework module

srri-zz:
Thanks for testing, ship it!

dispatch/settings.py
```diff
@@ -20,7 +23,7 @@
 SECRET_KEY = '&t7b#38ncrab5lmpe#pe#41coa-8ctwuy@tm0!x8*n_r38x_m*'
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = False
+DEBUG = True
```
srri-zz, line 26: This outputs a ton of junk per API call, just be aware..

apps/api/serializers.py
psiemens, line 14: Yeah, I only removed them because they were causing the route to fail. get_user_model() was spitting out the new User model I created.

dispatch/settings.py
psiemens, line 26: Huh, it wasn't outputting any junk for me. All it did was kill the CSS files. Maybe my setup is different?

PR #25, opened 2014-12-12
Add API for Articles and Resources

srri-zz:
Addition of:
- API endpoint for `articles` and `resources` models
- Added new requirement for mysqldb (python)
- Changed some imports to reflect the new structure

### DIFF
--- a/apps/api/serializers.py
+++ b/apps/api/serializers.py
@@ -1,12 +1,24 @@
 __author__ = 'Steven Richards'
 from django.contrib.auth.models import Group
 from django.contrib.auth import get_user_model
+from dispatch.apps.content.models import Resource, Article
 from rest_framework import serializers
 
-#class ArticleSerializer(serializers.HyperlinkedModelSerializer):
-#   class Meta:
-#        model = Article
-#        fields = ('headline')
+class ResourceSerializer(serializers.HyperlinkedModelSerializer):
+   class Meta:
+        model = Resource
+        fields = ('created_at','updated_at')
+
+class ArticleSerializer(serializers.HyperlinkedModelSerializer):
+    class Meta:
+        model = Article
+        fields = ('long_headline',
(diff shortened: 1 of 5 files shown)


srri-zz:
@psiemens or @blx, thoughts?

apps/api/serializers.py
```diff
@@ -1,12 +1,24 @@
 __author__ = 'Steven Richards'
 from django.contrib.auth.models import Group
 from django.contrib.auth import get_user_model
+from dispatch.apps.content.models import Resource, Article
```
psiemens, line 4: I believe this import statement can be simplified like so:

`from content.models import Resource, Article`

Django autoloads all the apps so you don't have to include the `dispatch.apps` part of the import statement, just the app name. Give it a try and see if it works.

psiemens:
I don't think we need an API route for Resource because it's just a base model. Am I right, @blx?

The mysqldb requirement is a good call.

blx:
Yeah it's abstract, I'm not sure it needs to be exposed in the API. As long as `Article` properly inherits the `Resource` fields.

Should `django` itself be in requirements? (If I make a new virtualenv and just install djangorestframework it doesn't pull down django.)

apps/api/serializers.py
blx, line 4: On the other hand, we could keep the fully qualified name to be explicit about what `content` or `core` etc are.

psiemens:
Yeah, I guess it should be.

apps/api/serializers.py
psiemens, line 4: Yeah, that might be best. Good to avoid naming clashes with external libs

srri-zz:
Fixed the route for resource, I wasn't sure what is was in terms of our modelling, thanks for clarifying. Also added django as a requirement. Merging since I've made those changes

PR #27, opened 2014-12-12
User models + admin

psiemens:
I made some changes to the structure for User and Person. I took out Contributor because I realized it was redundant. A User now must be linked to a Person, but a Person doesn't have to be a User, and a Person can have any number of ContributorRoles. 

user-person-contrib

I also changed the Resource model to make the author field associate with a Person instead of a User. This way we can set authors who aren't users.


apps/content/models.py
```diff
@@ -21,7 +19,7 @@ class Article(Resource):
     long_headline = CharField(max_length=200)
     short_headline = CharField(max_length=100)
     section = ForeignKey('Section')
-#   author = ForeignKey('User')     # TODO
+    author = ForeignKey(Person)
```
srri-zz, line 22: TODO: Update the `Article` API to support the `author` field now that is has been added

apps/core/admin.py
```diff
@@ -1,7 +1,74 @@
...
+        password2 = self.cleaned_data.get("password2")
+        if password1 and password2 and password1 != password2:
+            raise forms.ValidationError("Passwords don't match")
+        return password2
+
+    def save(self, commit=True):
+        # Save the provided password in hashed format
+        user = super(UserCreationForm, self).save(commit=False)
+        user.set_password(self.cleaned_data["password1"])
+        if commit:
```
srri-zz, line 32: Where is `cleaned_data` created? I don't see any references aside from where you call it.

apps/core/models.py
```diff
@@ -25,32 +25,37 @@ def create_superuser(self, email, password):
...
         return u'%s' % (self.title)
 
-class Contributor(Model):
-    person = OneToOneField(Person)
-    type = ManyToManyField(ContributorType)
+class Person(Model):
+    first_name = CharField(max_length=255)
+    last_name = CharField(max_length=255)
+    user = ForeignKey('User', blank=True, null=True, related_name='user')
+    roles = ManyToManyField(ContributorRole, blank=True, null=True)
```
srri-zz, line 38: TODO: Update API for additional `Person` fields

srri-zz:
Looks good for the most part, please see my comments though.

apps/core/admin.py
psiemens, line 32: The class extends django's ModelForm, which has methods/attributes that deal with sanitization. `cleaned_data` is generated automatically.

srri-zz:
@psiemens so we want to only expose `user` and not `person`?

psiemens:
@Captainkrtek I would still expose both for now.

srri-zz:
Okay, going to add that and make a PR

PR #28, opened 2014-12-12
playing around with modular theme support

psiemens:
This doesn't have to be merged yet -- I just wanted to let you guys know what I started working on. I want to give users the ability to easily build custom templates and swap them out for different ones. I made a themes package with a default theme inside it. I picture a "theme" being a collection of url routes, views, and templates. 

Creating a custom theme could be done in two ways:
1. Create a "template only theme" and use Dispatch's default routes and view.
2. Create a "full theme" with your own routes, views, and templates.

If a user chose option 2, they could still import specific routes and views from the default theme. So if somebody just wanted to add a few new types of pages they could do a partial override. 

I created a new model for system-wide settings that I'm using to store the current_theme variable. If no current_theme is set in the database, then Dispatch just picks the default theme. The reason I'm using a database to store this setting is so that users can easily select the default theme from the admin panel. 

Thoughts? @Captainkrtek @blx


srri-zz:
I like the idea but maybe we should hold off implementing it until the next iteration when things are a bit more fleshed out. Looks good though

psiemens:
Yeah, it doesn't have to happen right away. Just something I'm trying out.

psiemens:
Can we merge this into the main development cycle now? I think we're ready for it. I'll resolve conflicts locally and make sure Travis still passes it.

srri-zz:
Yea I'll merge it once it is able to be merged by Github

PR #29, opened 2014-12-13
cleaning up API and exposing author field

srri-zz:
A bit of clean up on the API with the modified Article model, removed groups for now.


psiemens:
Looks good to me, go ahead and merge.

Issue #30, opened 2014-12-13
Bug: ContributorRole conflict


srri-zz:
```
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute
    django.setup()
  File "/usr/local/lib/python2.7/site-packages/django/__init__.py", line 21, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python2.7/site-packages/django/apps/registry.py", line 108, in populate
    app_config.import_models(all_models)
  File "/usr/local/lib/python2.7/site-packages/django/apps/config.py", line 202, in import_models
    self.models_module = import_module(models_module_name)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/Users/krtek/Dev/dispatch/apps/core/models.py", line 28, in <module>
    class ContributorRole(Model):
  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 285, in __new__
    new_class._meta.apps.register_model(new_class._meta.app_label, new_class)
  File "/usr/local/lib/python2.7/site-packages/django/apps/registry.py", line 213, in register_model
    (model_name, app_label, app_models[model_name], model))
RuntimeError: Conflicting 'contributorrole' models in application 'core': <class 'core.models.ContributorRole'> and <class 'apps.core.models.ContributorRole'>.
```

This happens on the latest build of the `develop` branch when testing core

srri-zz:
To make this problem less likely to happen down the road we should consolidate models, or at least make it clearer. Cause right now we have users in two models I think and the namespaces might conflict/more likely to make a mistake.

psiemens:
I think I just fixed the problem. Sorry for the direct commit but it was only one line.

srri-zz:
It's all good! thanks! I'll try again

srri-zz:
All clear!

PR #31, opened 2014-12-13
Password hashing and API key stub

srri-zz:
We don't have to merge this if it doesn't make sense. I was unclear about if we are or are not hashing password. If not this should work. Couldn't test `core` because there is bug #30 @psiemens

### DIFF
--- a/apps/api/generator.py
+++ b/apps/api/generator.py
@@ -1,0 +1,18 @@
+__author__ = 'Steven Richards'
+
+from time import time
+from hashlib import sha512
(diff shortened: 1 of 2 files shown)


psiemens:
Django has hashing built in (it's automatic). Do you think this is better?

srri-zz:
Perhaps, I'll do more research but I think we should have a modern/secure implementation not just standard options django supplies

psiemens:
Sweet -- let me know what you find.

PR #32, opened 2014-12-13
expose persons model

srri-zz:
From comment on user and person models earlier, exposing the user model to the API. I don't think we will keep this here long term, just for playing with the models for now. @psiemens


psiemens:
Looks good.

PR #221, opened 2015-09-27
Fix headlines in search

jordanschalm:
Added safe filter to h2.article in search view to prevent html tags from being escaped and appearing in results


psiemens:
Looks good. Deployed.

PR #222, opened 2015-09-27
Searchbar: Autofocus on <input>, close with ESC

blx:


blx:
**Waiting to merge this** into dev until PR #221 gets merged into prod, otherwise these commits will tag along to prod with #221.

That would be fine, we'd just have to broaden the title of #221 so it also describes these changes.

PR #228, opened 2015-09-28
Use proper apostrophe in Ubyssey tagline

blx:
[From afar] But ben, when will you do something useful


blx:
i'll have you all know i did wait through a full 52-second travis build before merging this

PR #232, opened 2015-09-29
README enhancements

blx:
Mainly re: how to build static assets


blx:
Should've tagged this with `[ci skip]`

PR #235, opened 2015-09-29
Fix #233 where adblock breaks searchbar.

### LINKED ISSUE #233: Search button does not work on article pages
Pressing the search button while on an article does not show the search input.

Firefox 40.0.3.

blx:
This is hackish and we should look more into the effects of adblock etc, but in the meantime this fixes search.


psiemens:
Ads are broken in general on the article page anyways, so I'm not too worried about pushing a quick fix. `refreshAdvertisements` was an attempt to reload the ads after each AJAX call, but it doesn't seem to be working.

PR #241, opened 2015-09-30
Added controller for embedded code

jordanschalm:


psiemens:
Made a few modifications. Go ahead and commit/deploy.

dispatch/apps/frontend/embeds.py
```diff
@@ -84,7 +84,23 @@ def render(data):
...
 
+class CodeController:
+    @staticmethod
+    def json(data):
+        return data
+        
+    @staticmethod
+    def render(data):
+        if data['mode'] == 'css':
+            return '<style type="text/%s">%s</style>' % (data['mode'], data['content'])
```
blx, line 95: Do we want to bother specifying `type` here? On the one hand, it's explicit, but also superfluous.

`type` for `<style>` defaults to `text/css` in HTML5, which is the doctype and element set (eg. `<header>`) we're using. No browser is going to assume a different `type`.

Same thing for `text/javascript` below. Google.ca and Mozilla.org (for example) have lots of bare `<script>` and `<style>` tags.

PR #261, opened 2015-10-02
Fix #252 (dropdown toggle) and use relative paths in gulpfile

### LINKED ISSUE #252: Dropdown menus in article editor are not dismissed if they are visible when clicked

blx:


psiemens:
Nice work. Merging.

PR #269, opened 2015-10-06
Updated featured image updating code

jordanschalm:
This should fix errors when featured_image isn't yet defined.

@psiemens could you check on this and make sure it looks ok? I'm not 100% what feat_image should look like here.


psiemens:
Yep, this works. There probably _is_ a more elegant solution using `getDefaultProps()`, but this is fine in my opinion.

PR #275, opened 2015-10-06
Making custom credits work the way they should, moving to version 0.2.0

psiemens:


psiemens:
I've tested this on the staging server and it seems to be working. @jordanschalm can you take a look before we merge? Don't want a repeat of last time :grimacing:

jordanschalm:
This all looks good to me

PR #288, opened 2015-10-08
Cache search results on client for duration of page visit

blx:
Reduces number of search API calls with a transient client-side LRU cache of previously searched results.

This way if you're typing in the searchbar and then backspace to edit, we won't fire another query until it's actually different.


jordanschalm:
:+1:

Issue #215, opened 2015-09-27
Bug: no relative date formatting for ajax-loaded articles on a Section page


blx:
Go to [ubyssey.ca/features/](http://ubyssey.ca/features/) and scroll until the first set of ajax-loaded articles loads.

**Date-time looks like this `2015-09-07T13:19:26Z` instead of `Sept. 7, 2015, 1:23 p.m.`**

I think this is rendered in [Section.jsx#66](https://github.com/ubyssey/dispatch/blob/production/dispatch/themes/ubyssey/static/js/src/components/Section.jsx#L66) but idk where the date formatting code is.

jordanschalm:
Also ajax-loaded articles don't show special characters in snippets

jordanschalm:
The rendering code is the same in Section.jsx and in the template HTML files (just { article.published_at }). It looks like published_at gets assigned as a datetime object in [content/models.py](https://github.com/ubyssey/dispatch/blob/production/dispatch/apps/content/models.py#L198). I'm not sure when the datetime object gets converted to a string

PR #298, opened 2015-10-10
Fix issue #215 where AJAX-loaded articles in section views would have UTC-…

jordanschalm:
…formatted dates.


jordanschalm:
This is a bit of a hacky fix, let me know if anyone has a better way to fix this

PR #307, opened 2015-10-15
Only close selected dropdown if multiple are open

blx:
Minor tweak to dropdown behaviour in admin/manager.

I think this fixes
https://github.com/ubyssey/dispatch/commit/cf798960ceb53efa4ab39d442d699e96a5eca6a8#commitcomment-13738183.
The problem was that if more than one dropdown was open, selecting an
item in dropdown A would close all other dropdowns open, instead of just
closing A. Now a click event on dropdown A will not bubble beyond A.


jordanschalm:
Looks good :+1:

PR #375, opened 2015-12-03
Fix component editor text field switching issue

jordanschalm:


jordanschalm:
Also fixes safari footer layout issue

PR #413, opened 2016-03-27
updating admin to be more mobile-friendly

willmcdonald:
Updated admin with some media queries to make it functional on mobile


psiemens:
The styling looks good, but unfortunately you only updated the minified CSS file and not the source files themselves. We're using a CSS preprocessor called SASS that lets us write our CSS in a more organized, programmatic way. The file you edited is the compressed output that SASS generates.

All of the source files live in `css/sass`. I think most of the changes you need to make would be in this file: https://github.com/ubyssey/dispatch/blob/develop/dispatch/apps/manager/static/manager/css/sass/modules/_grid.scss

Here's what you need to do to compile the output CSS:
1. Download and install node - https://nodejs.org/en/download/
2. Open a terminal and cd to the `dispatch/apps/manager/static/manager/` directory on your computer.
3. Run `npm install` (might need to sudo it)
4. Then run `gulp sass` whenever you want to compile the CSS.

PR #437, opened 2017-01-22
everything now at least links to a dummy page

JimFung:
also added a space for the items in the drop down so its not right beside the icon


psiemens:
Looks great, thanks! Merging

PR #450, opened 2017-02-09
Beginning of Facebook Instant Articles integration

psiemens:
Adds support for:

- Dispatch integrations (FB Instant Articles is the only one so far, will probably add Slack, Twitter in the future)
- Management page for FB Instant Articles
- `integrations/` API endpoints
- OAuth authentication with Facebook Graph API
- Backend test suite! (will add more info about this later)

@ubyssey/dispatch It'd be awesome if you guys could look over this code and provide feedback. If you've got time, try running it locally and testing the Facebook authentication!

Before you try it out locally, run these commands to update your database schema:

```
python manage.py makemigrations core
python manage.py migrate core
```

To test Facebook authentication, first go to http://developers.facebook.com and create a new app (you can call it something like "Dispatch Test"). 

In the app settings page, add a website and set `http://localhost:8000/` as the URL.

You should then be able to use the App ID and App Secret to test the authentication flow.


dispatch/apps/api/serializers.py
```diff
@@ -437,3 +437,17 @@ def update(self, instance, validated_data, action='update'):
...
         return instance
+
+class IntegrationSerializer(serializers.Serializer):
+
+    id = serializers.CharField(source='ID', read_only=True)
+    settings = serializers.JSONField(source='get_settings')
+
+    def save(self):
+
+        settings = self.validated_data.get('get_settings', None)
```
blx: None is the default for `dict.get(nonexistent_key)`. Are you being explicit on purpose?

dispatch/apps/api/views.py
```diff
@@ -577,6 +578,71 @@ def retrieve(self, request, pk=None):
...
+    permission_classes = (IsAuthenticated,)
+    serializer_class = IntegrationSerializer
+
+    def get_object_or_404(self, pk=None):
+        try:
+            return integrationLib.get(pk)
+        except IntegrationNotFound:
+            raise NotFound('That integration does not exist')
+
+    def get_paginated_response(seld, data):
```
blx: > `seld`

👀

dispatch/apps/api/serializers.py
psiemens: Ah no that's a mistake

dispatch/apps/api/views.py
```diff
@@ -577,6 +578,71 @@ def retrieve(self, request, pk=None):
...
+
+        return Response(serializer.data)
+
+    def partial_update(self, request, pk=None):
+
+        integration = self.get_object_or_404(pk)
+
+        serializer = self.get_serializer(integration, data=request.data)
+
+        serializer.is_valid()
```
blx: I think as this is written it's a no-op. `is_valid()` returns a bool unless you pass the `raise_exception=True` flag

http://www.django-rest-framework.org/api-guide/serializers/#validation

dispatch/apps/api/views.py
psiemens: 😛

dispatch/apps/core/managers.py
```diff
@@ -0,0 +1,65 @@
...
+        return user
+
+    def create_user(self, email, password=None):
+        return self._create_user(email, password)
+
+    def create_superuser(self, email, password):
+        return self._create_user(email, password, True, True, True)
+
+    def is_valid_password(self, password):
+        return len(password) >= 8
```
blx: Is this intended to throw when password is none?

dispatch/apps/core/managers.py
```diff
@@ -0,0 +1,65 @@
...
+
+class IntegrationManager(Manager):
+
+    def get_settings(self, integration_id):
+        """Return settings for given integration as a dictionary."""
+
+        try:
+            integration = self.get(integration_id=integration_id)
+            return json.loads(integration.settings)
+        except self.model.DoesNotExist, ValueError:
```
blx: I think parens are required here if we want to strive for python3 compat
```python
except (self.model.DoesNotExist, ValueError):

File: dispatch/static/manager/src/js/components/SidebarNav.js [hunk 6]
```diff
@@ -0,0 +1,20 @@
...
+
+export default function SidebarNav(props) {
+
+  let items = props.items.map((item, i) => {
+    return (
+      <li className='c-sidebar-nav__link' key={i}>
+        <Link to={item.path}>{item.label}</Link>
+      </li>
+    )
+  })
```
blx: explicit return preferred over implicit?
```js
const items = props.items.map((item, i) => (
  <li className=...
  </li>
))

File: dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js [hunk 7]
```diff
@@ -0,0 +1,262 @@
+import React from 'react'
+import R from 'ramda'
+import { connect } from 'react-redux'
+
+import { FormSection, FormInput, TextInput, SelectInput } from '../../components/inputs'
+import { AnchorButton, Intent } from '@blueprintjs/core'
+
+import * as integrationActions from '../../actions/IntegrationActions'
+
+const DISPTACH_REDIRECT_URI = 'http://localhost:8000/admin/integrations/fb-instant-articles/?callback=1'
```
blx: is this always localhost?

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
```diff
@@ -0,0 +1,262 @@
...
+
+function fbLoginURI(clientId) {
+  return `https://www.facebook.com/v2.8/dialog/oauth?client_id=${clientId}&redirect_uri=${DISPTACH_REDIRECT_URI}&scope=pages_manage_instant_articles,pages_show_list`
```
blx: > DISPTACH_REDIRE...

sp. ?

psiemens: lol :man_facepalming:

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
```diff
@@ -0,0 +1,262 @@
...
+        'fb-instant-articles',
+        this.props.location.query
+      );
+    }
+
+  }
+
+  componentWillReceiveProps(nextProps) {
+
+    if (nextProps.integration && nextProps.integration.settings && nextProps.integration.settings.client_configured) {
```
blx:
```js
if (R.path(['integration', 'settings', 'client_configured'], nextProps)) {
```
if ya want

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
```diff
@@ -0,0 +1,262 @@
...
+    let pageMap = {}
+
+    let options = integration.callback.pages.data.map(page => {
+
+      // Add access token to map
+      pageMap[page.id] = page
+
+      return {
+        'value': page.id,
+        'label': page.name
```
blx: quoting keys on purpose?

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
```diff
@@ -0,0 +1,262 @@
...
+          {this.state.editMode ? cancelButton : null}
+        </FormInput>
+      </div>
+    )
+
+  }
+
+  render() {
+
+    let integration = this.props.integration
```
blx: is there a dispatch styleguide on let vs const? 80% of the `let`s i've seen here are effectively immutable and could be marked as such if you were so inclined

PR #451, opened 2017-02-13
Dashboard Page

JimFung:
Completes #451 

A couple things to pay attention to: 
1) my css is probably trash, please let me know how I can make it better
2) following up on point 1, I have no idea how to make the list of user actions wrap nicely. open to ideas. (see attached photo)
3) right now the links in the list of actions link directly to the article itself instead of the editor. Not sure if that was the intention or not. 
4) I kept everything inside dashboardpage.js but i don't mind breaking it out into separate components. 
5) Don't think there is a "New Page" action yet, so it doesn't link to anything right now.

screen shot 2017-02-12 at 5 34 37 pm


dispatch/apps/core/actions.py
```diff
@@ -40,16 +40,18 @@ def list_actions(count=25):
...
                 article = Article.objects.get(parent_id=action.object_id, head=True)
-                if count == 1:
-                    text = '%s %s <a href="%s">%s</a>' % (action.person.full_name, SINGULAR[action.action], article.get_absolute_url(), article.headline)
-                else:
-                    text = '%s made %d %s to <a href="%s">%s</a>' % (action.person.full_name, count, PLURAL[action.action], article.get_absolute_url(), article.headline)
+                meta = {}
+                meta['author'] = action.person.full_name
+                meta['headline'] = article.headline
+                meta['article_url'] = article.get_absolute_url()
+                meta['count'] = 1 if count == 1 else count
```
blx: isn't this equivalent to
```py
meta['count'] = count

File: dispatch/static/manager/src/styles/components/dashboard.scss [hunk 2]
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.dashboard_container h2 {
+  font-size: 1.5em;
+  font-weight: bolder;
+}
+
+.dashboard_container ul li {
+
+}
```
blx: empty style?

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
+    //returns the user history if the data is loaded, else returns an empty list element
+    let historyList = this.props.actions.isLoaded
+    ? this.props.actions.data.map((elem,ind) => (
+      <li key={ind}>
+        <i className={`fa fa-${elem.icon}`} aria-hidden="true"></i>
+        {elem.meta.author}
+        {elem.meta.count == 1 ? ` ${elem.meta.action} ` : ` made ${elem.meta.count} ${elem.meta.action} to ` }
+        <a href={elem.meta.article_url}>{elem.meta.headline}</a>
+        <span>{` ${moment(elem.timestamp).from(moment())}`}</span>
+      </li>))
```
blx: re point # 2 -- one approach would be to float the icon left, put the text in an inline-block or block element, and slap a margin-left on it. it should wrap at the same indent:

screen shot 2017-02-12 at 9 51 39 pm

```html
    <style>
        .a { margin-left: 2rem; background: #ddd; max-width: 25rem; margin-top: 1rem; }
        i { float: left; }
    </style>

    <div>
        <i class="fa fa-server"></i>
        <div class="a">It was a bright cold day in April, and the clocks were striking thirteen.</div>
    </div>
    <div>
        <i class="fa fa-feed"></i>
        <div class="a">Winston Smith, his chin nuzzled into his breast in an effort to escape the vile wind, slipped quickly through the glass doors of Victory Mansions, though not quickly enough to prevent a swirl of gritty dust from entering along with him.</div>
    </div>
```
(this is divs, not `li`s, but same principle)

dispatch/static/manager/package.json
```diff
@@ -54,5 +55,8 @@
   },
   "bugs": {
     "url": "https://github.com/ubyssey/dispatch/issues"
+  },
+  "dependencies": {
+    "moment": "^2.17.1"
```
blx: is moment the only non dev dependency? why is the min version here different from on line 30 above?

dispatch/apps/core/actions.py
JimFung: why yes, yes it is. 😅

dispatch/static/manager/src/styles/components/dashboard.scss
JimFung: forgot to delete this.

dispatch/static/manager/package.json
JimFung: hmm not sure how it got into devDependencies since i ran npm with the --save tag. I didn't realize our dependencies were all devDependencies, is there any reason why that is? I believe the convention is to save anything required for our app to function as a dependency so when we install it in a production env (with the --production flag for npm) it'll ignore the devDependencies, which are generally linters, test runners, etc. 

I will remove moment from dependencies and leave the one in the devDependency object, since that is the convention this project is using.

dispatch/static/manager/src/js/pages/DashboardPage.js
JimFung: Thanks, I'll give this a shot!

dispatch/static/manager/package.json
blx: yeah i'd have the same reasoning ✌️ maybe ping @psiemens on this

dispatch/static/manager/src/js/pages/DashboardPage.js
JimFung: I tried this but it still didn't wrap nicely 😕 maybe i didn't do it right. I'll try it again tonight, if not i might just stick these inside a table cause that'll work... :crossed_fingers:

JimFung: image
Nailed it...kinda, that one fifth from the bottom still doesn't look great but it wraps nicely.

dispatch/apps/core/actions.py
```diff
@@ -40,16 +40,18 @@ def list_actions(count=25):
...
-                if count == 1:
-                    text = '%s %s <a href="%s">%s</a>' % (action.person.full_name, SINGULAR[action.action], article.get_absolute_url(), article.headline)
-                else:
-                    text = '%s made %d %s to <a href="%s">%s</a>' % (action.person.full_name, count, PLURAL[action.action], article.get_absolute_url(), article.headline)
+                meta = {}
+                meta['author'] = action.person.full_name
+                meta['headline'] = article.headline
+                meta['article_url'] = article.get_absolute_url()
+                meta['count'] = count
+                meta['action'] = SINGULAR[action.action] if count == 1 else PLURAL[action.action]
```
psiemens: I'd prefer to write it like this:

```python
meta = {
    'author': action.person.full_name,
    'headline': article.headline,
    ...etc..
}
```
Just more readable IMO.

dispatch/static/manager/package.json
psiemens: If I recall correctly, there used to be a time when saving deps as `devDependencies` was the norm for Webpack projects (since all of them are bundled anyways), but I think that's changed. I'm cool with moving everything over to `dependencies`.

dispatch/static/manager/src/js/actions/DashboardActions.js
```diff
@@ -0,0 +1,24 @@
...
+    payload: DispatchAPI.dashboard.actions(token)
+      .then(json => {
+        return json
+      })
+  }
+}
+
+export function getRecentArticles(token) {
+  return {
+    type: types.FETCH_RECENT,
```
psiemens: Can we make this a bit more verbose?

```js
types.FETCH_RECENT_ARTICLES
```

dispatch/static/manager/src/js/api/dispatch.js
```diff
@@ -155,6 +155,14 @@ var DispatchAPI = {
     createTag: (token, name) => {
       return postRequest('tags', null, {name: name}, token)
     }
+  },
+  dashboard: {
```
psiemens, line 159: We're sort of breaking out of REST here, but I'm not totally opposed to that...

I'm just wondering if it then makes sense to combine both endpoints, or at least put them under the same root:

`GET api/dashboard/actions`
`GET api/dashboard/recent`

Thoughts?

dispatch/static/manager/src/js/pages/DashboardPage.js
psiemens: I played around with this a little bit and got it working nicely with a flexbox. If you don't mind, I can push a fix.

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
+
+  componentWillMount() {
+    this.props.fetchActions(this.props.token)
+    this.props.fetchRecent(this.props.token)
+  }
 
   render() {
+
+    //returns the user history if the data is loaded, else returns an empty list element
+    let historyList = this.props.actions.isLoaded
```
psiemens: This can be a `const`

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
+        <i className={`fa fa-${elem.icon}`} aria-hidden="true"></i>
+        {elem.meta.author}
+        {elem.meta.count == 1 ? ` ${elem.meta.action} ` : ` made ${elem.meta.count} ${elem.meta.action} to ` }
+        <a href={elem.meta.article_url}>{elem.meta.headline}</a>
+        <span>{` ${moment(elem.timestamp).from(moment())}`}</span>
+      </li>))
+    : <li></li>
+
+    //Returns a list of the user's recent articles if the data is loaded, else returns an empty list element
+    let recentArticlesList = this.props.recent.isLoaded
```
psiemens: Change to `const`

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
+    ? this.props.recent.data.map((elem,ind) => {
+      return (
```
psiemens: Change this to implicit return like above

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
 import React from 'react';
 import DocumentTitle from 'react-document-title';
+import { Link } from 'react-router'
+import { connect } from 'react-redux'
+import * as dashboardActions from '../actions/DashboardActions'
+import moment from 'moment'
```
psiemens: This is nit-picky, but I like to put all external imports before project imports. Can you move this up one line?

dispatch/static/manager/src/js/reducers/DashboardReducer.js
```diff
@@ -0,0 +1,61 @@
...
+    data: []
+  },
+  recent: {
+    isLoading: false,
+    isLoaded: false,
+    data: []
+  }
+}
+
+let actionsReducer = (state = initialState.actions, action) => {
```
psiemens: Change to `const`

dispatch/static/manager/src/js/reducers/DashboardReducer.js
```diff
@@ -0,0 +1,61 @@
...
+        isLoading: false,
+        isLoaded: true,
+        data: action.payload.results
+      })
+    default:
+      return state
+  }
+}
+
+let recentReducer = (state = initialState.recent, action) => {
```
psiemens: Change to `const`

dispatch/static/manager/src/styles/components/_dashboard.scss
```diff
@@ -0,0 +1,61 @@
+.dashboard_container {
```
psiemens: Check out the SASS style guide here: https://github.com/ubyssey/ubyssey-dispatch-theme/wiki/SASS-Style-Guide

Sorry, should have let you know before. Shouldn't be hard to change these styles though

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
+            <h2>Latest Activity</h2>
+            <ul className="dashboard_user_history">
+              {historyList}
+            </ul>
+          </div>
+          <div className="dashboard_right">
+            <h2>Quick Actions</h2>
+            <div className="dashboard_quick_actions">
+              <ul>
+                <li><button className="pt-button"><Link to={`/articles/new`}>New Article</Link></button></li>
```
psiemens: There's actually a component for this!

```js
import { LinkButton } from '../components/inputs'

...

<LinkButton to='articles/new'>New Article</LinkButton>
```

dispatch/static/manager/src/js/pages/DashboardPage.js
```diff
@@ -1,14 +1,85 @@
...
   render() {
+
+    //returns the user history if the data is loaded, else returns an empty list element
+    let historyList = this.props.actions.isLoaded
+    ? this.props.actions.data.map((elem,ind) => (
+      <li key={ind}>
+        <i className={`fa fa-${elem.icon}`} aria-hidden="true"></i>
+        {elem.meta.author}
+        {elem.meta.count == 1 ? ` ${elem.meta.action} ` : ` made ${elem.meta.count} ${elem.meta.action} to ` }
+        <a href={elem.meta.article_url}>{elem.meta.headline}</a>
```
psiemens: Can we change this to link to the article edit page in Dispatch instead of the live version?

You should be able to do something like this:

```js
<LinkButton to={`articles/${elem.meta.article_id}`}>{elem.meta.headline}</LinkButton>
```

You'll just have to add the article ID to the API response first :metal:

psiemens (changes requested):
Everything looks really good for the most part! The functionality is nearly perfect.

dispatch/apps/core/actions.py
JimFung: 👍 was hoping for more review on my python cause i don't know the syntax at all. Changing this now.

dispatch/static/manager/src/js/pages/DashboardPage.js
JimFung: 👍 yes please do! i wanna see how you did it.

dispatch/static/manager/src/js/api/dispatch.js
JimFung: Honestly didn't give it much thought when I built it, figured it'd be easy to move around once the functionality was there. 

/{page}/{data} is probably the most common pattern and i'm not opposed to it. 

Out of habit, I make my endpoints cause this is how i designed my APIs at work cause its a good way to dig through large nested datasets. 
/{resource}/{id/some_other_dimension}

I think for your suggestion is probably better here so i'll change it to that. 👍

dispatch/static/manager/src/js/pages/DashboardPage.js
JimFung: yep, i'll make the api change as part of the change to make the endpoints /dashboard/{resource}

dispatch/static/manager/src/js/api/dispatch.js
JimFung: Done!

psiemens (approved):

psiemens:
Just made a few updates to the CSS and JSX code. Merging!

PR #452, opened 2017-02-13
Logout

JimFung:
Completes #453 
But more importantly the padding on the delete button is fixed. 

Logout removes the JWT and email from the user's cookies but doesn't remove the token from the server.


dispatch/static/manager/src/js/index.js
```diff
@@ -53,6 +53,8 @@ render((
           <Route path='/login' component={Pages.Login} />
         </Route>
 
+        <Route path='/logout' component={Pages.Logout} />
```
JimFung, line 75: I think i put this in the right place, if it should be another level deeper or w.e lmk but i think this looks right.

JimFung:
TODO: remove token from server

JimFung:
Current plan: 
I want the /auth/token endpoint to manage both login and logout. To achieve this I'll be refactoring the login/logout code so that it'll be in its own viewset. If there is a better way to do this, Give me a shout but this is the direction I'll be going forward with right now.

dispatch/apps/api/urls.py
```diff
@@ -34,6 +34,11 @@
 article_bulk_delete = views.ArticleViewSet.as_view({ 'post': 'bulk_delete' })
 article_comments = views.CommentViewSet.as_view({ 'get': 'article' })
 
+authentication_management = views.AuthenticationViewSet.as_view({
+    'post': 'user_authenticate',
+    'delete': 'user_unauthenticate'
+})
```
psiemens, line 45: Dope! This is exactly how the `as_view` method is supposed to be used :metal:

dispatch/apps/api/urls.py
```diff
@@ -48,5 +53,5 @@
     url(r'^articles/delete/$', article_bulk_delete, name='article-bulk-delete'),
     url(r'^articles/(?P<pk>[0-9]+)/comments/$', article_comments, name='article-comments'),
     # User authorization
-    url(r'^auth/token', views.user_authenticate, name='user-token'),
+    url(r'^auth/token', authentication_management, name='authentication_management'),
```
psiemens: Can we change this to:

```python
url(r'^auth/token', authentication_management, name='auth-token'),

File: dispatch/apps/api/views.py [hunk 4]
```diff
@@ -638,24 +638,35 @@ def list(self, request):
...
-        return Response(data, status=status.HTTP_202_ACCEPTED)
-    else:
-        return Response({}, status=status.HTTP_400_BAD_REQUEST)
+            return Response(data, status=status.HTTP_202_ACCEPTED)
+        else:
+            return Response({}, status=status.HTTP_400_BAD_REQUEST)
+
+    def user_unauthenticate(self, request):
+
+        user_token = request.data
```
psiemens: Upon investigation, I don't think we need to be passing the token in the request payload. 

As long as the token header is set (as it already is), DRF will automatically handle authentication and we don't need to access the token string itself. 

I see what you're trying to do here, but I think we can simplify the entire method like so:

```python

def user_unauthenticate(self, request):

    # Get the token for the current user
    # In theory this should never 404, but doesn't hurt to include for safety
    token = get_object_or_404(Token, user=request.user)

    token.delete()

    return Response({}, status=status.HTTP_200_OK)

File: dispatch/static/manager/src/js/api/dispatch.js [hunk 5]
```diff
@@ -92,6 +105,9 @@ var DispatchAPI = {
       }
 
       return postRequest('auth/token', null, payload)
+    },
+    deleteToken: (token) => {
+      return delRequest('auth/token', null, token, token)
```
psiemens: If we make the backend change above, we don't need to include the token in the payload:

```python
return delRequest('auth/token', null, null, token)
```

dispatch/static/manager/src/js/pages/LogoutPage.js
```diff
@@ -0,0 +1,40 @@
...
+    this.props.Logout(this.props.token);
+  }
+
+  componentDidMount() {
+    // Assume it was successful if it reaches here, redirect to login page.
+    this.props.redirect();
+  }
+
+  render(){
+    return (<p>Hasta la vista, baby</p>)
```
psiemens: Bon voyage 👋

psiemens (changes requested):

dispatch/apps/api/views.py
```diff
@@ -638,24 +638,35 @@ def list(self, request):
...
+            data = {
+                'token': unicode(token)
+            }
 
-        return Response(data, status=status.HTTP_202_ACCEPTED)
-    else:
-        return Response({}, status=status.HTTP_400_BAD_REQUEST)
+            return Response(data, status=status.HTTP_202_ACCEPTED)
+        else:
+            return Response({}, status=status.HTTP_400_BAD_REQUEST)
```
blx: Recognising that this code was already here, so feel free to ignore:

- Do you think 403 would be a more semantic and standard response here? 400 generally corresponds to "you've sent malformed or bad data", whereas the case here seems more like the request was understood but the server is unwilling to accept it -- 403.
  - https://tools.ietf.org/html/rfc7231#section-6.5.1

- 400 would be appropriate if `email` or `password` or both were not sent at all. On that note, are there going to be cases where attempting to login with just an email or just a password is ok? If not, it would be better for lines 645-646 to look more like:
    ```py
    try:
        email = request.data.get('email')
        password = request.data.get('password')
    except KeyError:
        return Response({}, status=status.HTTP_400_BAD_REQUEST)

dispatch/apps/api/views.py
```diff
@@ -638,24 +638,35 @@ def list(self, request):
...
+
+        user_token = request.data
+        if user_token is None:
+            return Response({}, status=status.HTTP_400_BAD_REQUEST)
+        else:
+            try:
+                Token.objects.get(user_token).delete()
+            except:
+                pass
+            return Response({}, status=status.HTTP_200_OK)
```
blx: [200 is the default status](http://www.django-rest-framework.org/api-guide/responses/#response) if none is provided; I think it's more clear to omit 200s, so that the cases where you specify a different status are more obvious, but hey that's up to you.

dispatch/static/manager/src/js/actions/UserActions.js
```diff
@@ -46,3 +46,20 @@ export function authenticateUser(email, password, nextPath = '/') {
...
+    payload: DispatchAPI.articles.deleteArticles(token, articleIds)
+      .then( json => json.deleted )
+  }
+}
+
+export function unauthenticateUser(token) {
+  return {
+    type: types.AUTH_DELETE_TOKEN,
+    payload: DispatchAPI.auth.deleteToken(token)
+      .then( json => json )
```
blx: what does this accomplish? isn't this the identity function?

dispatch/apps/api/urls.py
JimFung: 🐵 see 🐵 do 🎉

dispatch/apps/api/views.py
JimFung: neat, thought i had to explicitly set the status code. I agree with what you said, gonna remove it.

dispatch/static/manager/src/js/actions/UserActions.js
JimFung: 🤔 I'll get back to you on that

JimFung: it is the identity function, what does it accomplish? it sparked a mildly interesting back and forth. 👍 taking it out since it's done its job.

dispatch/static/manager/src/js/pages/LogoutPage.js
JimFung: 👋

psiemens:
I fixed the issue of not being able to log back in immediately after a logout. It turns out that because we weren't creating a copy of the `DEFAULT_HEADERS` global object in `api/dispatch.js`, after the token had been set once it was being persisted until a page refresh. I fixed this by merging the `DEFAULT_HEADERS` object with the user-defined headers to create a fresh object every time.

However, I noticed another bug that I think is worth fixing. If a logout fails, the user is still directed to the login page and the logout page is never shown. After a bit of thought, I came to the conclusion that it doesn't really make sense to have `LogoutPage` at all. Instead, I think that `unauthenticateUser` should be passed as an action to `MainContainer`, which will then pass it onto `Header`. If a logout is successful, the user should be directed to the login page, and if not, an error message should be displayed. 

For an example of how to accomplish this with just actions, check out this code: https://github.com/ubyssey/dispatch/blob/logout/dispatch/static/manager/src/js/actions/UserActions.js#L35-L48

PR #476, opened 2017-02-24
Goodbye Gulp, hello Webpack & Yarn

psiemens:
New build process:

```bash
# Setup
yarn setup

# Development
yarn start

# Production
yarn build
```


dispatch/static/manager/package.json
```diff
@@ -7,45 +7,45 @@
...
-    "react-icons": "^2.2.1",
     "react-redux": "^4.4.5",
-    "react-redux-loading-bar": "^2.3.1",
+    "react-redux-loading-bar": "2.6.6",
     "react-router": "^2.7.0",
     "react-router-redux": "^4.0.5",
     "redux": "^3.5.2",
     "redux-promise-middleware": "^4.0.0",
     "redux-thunk": "^2.1.0",
+    "sass-loader": "^6.0.2",
```
JimFung: certain things like webpack loaders and webpack itself can be moved into a devDependency object.

dispatch/static/manager/src/js/index.js
```diff
@@ -26,7 +28,14 @@ const store = createStore(
     loadingBar: loadingBarReducer,
     modal: modalReducer
   }),
-  applyMiddleware(thunk, router, promiseMiddleware(), loadingBarMiddleware())
+  applyMiddleware(
+    thunk,
+    router,
+    promiseMiddleware(),
+    loadingBarMiddleware({
+      promiseTypeSuffixes: ['PENDING', 'FULFILLED', 'REJECTED'],
```
JimFung, line 36: Do we need to add an `${ACTION_TYPE}_REJECTED` case to all the reducers as part of this change?

dispatch/static/manager/package.json
```diff
@@ -54,5 +54,10 @@
   },
   "bugs": {
     "url": "https://github.com/ubyssey/dispatch/issues"
+  },
+  "scripts": {
+    "init": "yarn install --force",
```
JimFung: init is already a keyword for npm, i suggest changing it to something.

JimFung (commented):
I think this would also be a good chance to update the instructions in the README a bit. Maybe just adding which npm commands to run to get started to help lower the barrier of entry for new contributors who may not be as familiar with our process.

dispatch/static/manager/src/js/index.js
psiemens: Nope, those actions are generated automatically by the `promiseMiddleware`

JimFung:
addresses #477

dispatch/static/manager/src/js/index.js
JimFung: If it returned an action with the action type 'FETCH_ARTICLE_REJECTED' for example, we don't have a specific case set up to handle it in the articles reducer. My question is whether or not we want a specific case to handle this action.

dispatch/static/manager/package.json
psiemens: Good call.

psiemens: But technically every package is a devDependency when using Webpack...

dispatch/static/manager/src/js/index.js
psiemens: Ohhhh, I see. Yeah, that's probably something we should standardize 😛 . Let's issue it up

JimFung: Okay, I'll make an issue before the next meeting highlighting which ones need to be refactored. I'm going to exclude any reducer that needs to be changed as part of implementing a page/component that we already have an issue for (ie. not going to include the people reducer, for example since its already going to be changed when we implement the people page)

psiemens: @JimFung Sounds good!

dispatch/static/manager/package.json
JimFung: This distinction is also for when we install packages in a production setting, since packages under devDependency won't be install when we use the --production flag.

dispatch/static/manager/package.json
psiemens: Done

PR #478, opened 2017-02-27
Implement video embed

### LINKED ISSUE #455: Add video embeds

psiemens:
Close #455


dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+
+import { FormInput, TextInput } from '../../inputs'
+
+class VideoEmbedComponent extends React.Component {
+
+  constructor(props){
```
psiemens: As a convention, we like to leave a space before the opening bracket of a function:

```js
constructor(props) {
```

(Change this for the other methods too)

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+
+    )
+  }
+
+
+  render(){
+
+    return(
+      <div className='o-embed o-embed--image'>
+        {this.state.inserted ? this.renderVideo() : this.renderInput()}
```
psiemens: Instead of using `this.state.inserted` to track if a video has been inserted, you can just check to see if `this.props.data.id` has been set. This will fix the problem where "Insert" has to be clicked twice.

I think you can actually get rid of `this.state` completely and make this a stateless component. We don't need to include `validURL` in the state -- just don't set the id if it isn't valid.

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+
+    )
+  }
+
+
+  render(){
+
+    return(
+      <div className='o-embed o-embed--image'>
```
psiemens: Should be `<div className='o-embed o-embed--video'>`

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+        <button className="c-input--button"  onClick={this.insertVideo}>Insert</button>
+      </div>
+
+    )
+
+  }
+  renderVideo(){
+    return(
+      <div>
+        <img className='o-embed--image__image'src={"http://img.youtube.com/vi/" + this.props.data.id + "/0.jpg"}/>
```
psiemens: Change the className to video instead of image, and split onto multiple lines. You can also use a string template for the URL:

```jsx
<img 
  className='o-embed--video__image'
  src={`http://img.youtube.com/vi/${this.props.data.id}/0.jpg`} />

File: dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+      <div>
+        <form>
+          <FormInput label='URL'>
+            <TextInput
+              fill={true}
+              value={this.props.data.url}
+              onChange={this.handleUrlChange} />
+          </FormInput>
+        </form>
+        <button className="c-input--button"  onClick={this.insertVideo}>Insert</button>
```
psiemens: You can use the `Button` component here:

```jsx

import { Button } from '@blueprintjs/core'

...

<Button onClick={e => this.insertVideo()}>Insert</Button>
```

I also want to move away from binding methods in the constructor (it just looks ugly IMO). If you use the function shorthand as shown above, you can remove `this.insertVideo = this.insertVideo.bind(this)` from the constructor.

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  validYouTube(url){
+     var p = /^(?:https?:\/\/)?(?:www\.)?youtube\.com\/watch\?(?=.*v=((\w|-){11}))(?:\S+)?$/
+     return (url.match(p)) ? RegExp.$1 : false
+  }
+
+
+  renderInput(){
+    return(
```
psiemens: Add a space:

```jsx
return (
```

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+          </FormInput>
+        </form>
+        <button className="c-input--button"  onClick={this.insertVideo}>Insert</button>
+      </div>
+
+    )
+
+  }
+  renderVideo(){
+    return(
```
psiemens: Add a space:

```jsx
return (
```

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+      <div>
+        <img className='o-embed--image__image'src={"http://img.youtube.com/vi/" + this.props.data.id + "/0.jpg"}/>
+        <form>
+          <FormInput label='URL'>
+            <TextInput
+              fill={true}
+              value={this.props.data.url}
+              onChange={this.handleUrlChange} />
```
psiemens: You can do the same here as we did with the `insertVideo` method:

```jsx
onChange={e => this.handleUrlChange(e.target.value)}
```

psiemens (commented):
@JamieRL I left some comments that should address some of the problems we talked about. 

As for the editor not rendering videos after they have been saved, that's because the code to parse embeds from the API response isn't written yet (oops). So don't worry about that for now

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+    this.state = {
+      inserted: false,
+      validURL: false
+    }
+  }
+
+  handleUrlChange(e){
+    e.preventDefault()
+    this.props.updateField('url', e.target.value)
+    console.log("URL IS: "+this.props.data.url);
```
blx: is this, and the log on line 30, still being used?

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+      this.setState({
+        validURL: true,
+        inserted : true
+      })
+    }
+  }
+
+  validYouTube(url){
+     var p = /^(?:https?:\/\/)?(?:www\.)?youtube\.com\/watch\?(?=.*v=((\w|-){11}))(?:\S+)?$/
+     return (url.match(p)) ? RegExp.$1 : false
```
blx: `RegExp.$1..9` is [nonstandard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/n) and [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features#RegExp_Properties) notation, according to Mozilla, and refers to icky global state to boot.

It would be more reliable to use the array returned by [`url.match(p)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match). For example:

```js
const p = ...
const [, videoId] = url.match(p) || []
return videoId

File: dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js [hunk 11]
```diff
@@ -0,0 +1,95 @@
...
+
+    if(this.props.data.id){
+      this.setState({
+        validURL: true,
+        inserted : true
+      })
+    }
+  }
+
+  validYouTube(url){
```
blx: I think this name can be improved. `validX` sounds like a predicate: you would expect either `true` or `false` to be returned (valid, or not valid).

Something like `getYouTubeVideoId` or `getVideoId`, etc., would be much more clear to the reader, and probably to yourself in 3 months.

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+     var p = /^(?:https?:\/\/)?(?:www\.)?youtube\.com\/watch\?(?=.*v=((\w|-){11}))(?:\S+)?$/
```
blx: Is it deliberate that this does not accept YouTube short urls, which are fairly common? (eg. the "Share" panel gives you short urls, with the pattern `https://youtu.be/<ID>`).

example:
screen shot 2017-02-26 at 10 34 41 pm

JimFung: this could be a `const`

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
...
+  handleUrlChange(e){
+    e.preventDefault()
+    this.props.updateField('url', e.target.value)
+    console.log("URL IS: "+this.props.data.url);
+
+  }
+
+
+  insertVideo(){
+    var id = this.validYouTube(this.props.data.url);
```
JimFung: this could be a const

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+
+import { FormInput, TextInput } from '../../inputs'
+
+class VideoEmbedComponent extends React.Component {
+
+  constructor(props){
+    super(props)
+
+    this.handleUrlChange = this.handleUrlChange.bind(this)
```
JimFung: pretty sure you don't have to add the bind if you [wrote the functions using () => syntax ](https://medium.com/@machnicki/handle-events-in-react-with-arrow-functions-ede88184bbb#.2dlb31qok)

```
handleUrlChange = (e) => {
   ....
}
```

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
psiemens: Yup

JamieRL:
Ok I'm done vacationing. Will go through all this stuff today

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
JamieRL: Ideally it should accept both. I would guess that most people would just copy paste the full URL

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
JamieRL: I hadn't yet done final changes when the PR was made. Will remove all the logs etc.

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
JamieRL: Sounds good :)

dispatch/static/manager/src/js/components/ContentEditor/embeds/VideoEmbed.js
blx: ah ok cool :)

PR #482, opened 2017-02-28
Fixes undefined object in fetch_recent_articles_fulfilled

### LINKED ISSUE #481: fetch recent articles fulfilled will error if the current user has no recent articles
### Current behaviour
The following Error is thrown when you navigate to /admin **IF** you have no recent articles for the currently logged in user.
```
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
```

### Steps to reproduce 
1. Remove any article associated with the currently logged in user (if you're using a fresh dev db instance and the volunteer account, you're already good to go).
2. open the browser console
3. navigate to /admin

_Tested on the develop branch_

### Expected behaviour
No errors should be thrown, the reducer should gracefully handle the undefined and return an empty object.

### Additional Notes
...gdi

JimFung:
fixes #481.

Added a fallback to the Object.keys call so if no articles exist it will call Object.keys with an {} instead of undefined.

### DIFF
--- a/dispatch/static/manager/src/js/reducers/DashboardReducer.js
+++ b/dispatch/static/manager/src/js/reducers/DashboardReducer.js
@@ -40,7 +40,7 @@ const recentReducer = (state = initialState.recent, action) => {
         isLoading: true
       })
     case `${types.FETCH_RECENT_ARTICLES}_FULFILLED`:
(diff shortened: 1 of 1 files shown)


blx (approved):

JimFung:
thx for the review @blx 👍

PR #484, opened 2017-03-13
Setup testing framework

psiemens:
Yay for automated testing!

Test files live in `dispatch/tests` and should be prefixed with `test_`.

New commands:

```bash
# Run tests
dispatch-admin test

# Generate & display coverage report
dispatch-admin coverage

# Generate html coverage report
dispatch-admin coverage_html
```

@JamieRL you can use [this file](https://github.com/ubyssey/dispatch/blob/testing/dispatch/tests/test_api_articles.py) as an example when writing tests for the files endpoint.

Next steps:
    a) Write more tests! Let's start with all of the API routes and then move into internal modules
    b) Setup continuous integration


dispatch/bin/dispatch-admin
```diff
@@ -44,8 +49,51 @@ def start_project(name):
...
+
+    TestRunner = get_runner(settings)
+    test_runner = TestRunner()
+
+    failures = test_runner.run_tests(
+        [os.path.join(PACKAGE_DIR, 'tests')]
+    )
+
+    if failures:
+        sys.exit(bool(failures))
```
blx: Since you've already established that failures is truthy on the previous line, this would be more clear as

```py
sys.exit(1)

File: dispatch/tests/test_settings.py [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+AUTH_USER_MODEL = 'core.User'
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.sqlite3',
+        'NAME': 'test_dispatch',
+    }
+}
+
+REST_FRAMEWORK = {
```
blx: it might make sense to add a comment here and in the actual settings reminding to keep them in sync -- given that some tests are really just testing that the correct permissions are being applied.

psiemens: Maybe we should just do:

```python
# test_settings.py

from dispatch.conf.default_settings import REST_FRAMEWORK
```

dispatch/bin/dispatch-admin
```diff
@@ -44,8 +49,52 @@ def start_project(name):
...
+
+    # Run tests
+    run_tests()
+
+    # Stop tracking coverage
+    cov.stop()
+    cov.save()
+
+    # Show report
+    cov.report()
```
JimFung: This currently prints out the coverage in the terminal, but if we use html_report() we can have it print out the coverage report in .html. This is a bit more overhead but it does show us which lines are missing coverage. I'll add a screenshot below of what it looks like. 
screen shot 2017-03-13 at 2 41 37 pm

psiemens: I'll add a separate command for the html report!

PR #485, opened 2017-03-14
Updated itemlist to be more generic

JamieRL:
Changed itemlist to use passed in type, articles page now passes in articles as type


JamieRL:
Found a few more things that need to be changed. In itemlistitem we have fields for author and headline that can be abstracted

JimFung:
10/10 teamwork @JamieRL @psiemens

psiemens:
gucci

PR #486, opened 2017-03-15
Bug fix: links and images being removed after save

### LINKED ISSUE #445: Fix article editor to persist hyperlinks on save

psiemens:
Close #445


dispatch/static/manager/src/js/components/ContentEditor/ContentStateHelper.js
```diff
@@ -46,12 +30,71 @@ function parseBlock(block) {
...
+  return {
+    contentBlocks: [contentBlock],
+    entityMap: {entityKey: entity}
+  }
+
+}
+
+function createBlock(acc, block) {
+
+  var blocksFromJSON
```
JimFung: should be a let

PR #492, opened 2017-03-29
Improve Instant Articles integration page: split into separate components

psiemens:
Today's CPSC 310 lecture inspired me to improve this code. 

Tested with a sample Facebook App. plz merge 😄

### DIFF
--- a/dispatch/static/manager/src/js/components/integrations/FBInstantArticles.js
+++ b/dispatch/static/manager/src/js/components/integrations/FBInstantArticles.js
@@ -1,0 +1,141 @@
+import React from 'react'
+
+import { AnchorButton, Intent } from '@blueprintjs/core'
+
+import { FormInput, TextInput, SelectInput } from '../inputs'
+
+function fbLoginURI(clientId, redirectURI) {
+  return `https://www.facebook.com/v2.8/dialog/oauth?client_id=${clientId}&redirect_uri=${window.location.origin}/admin${redirectURI}?callback=1&scope=pages_manage_instant_articles,pages_show_list`
+}
+
+export function InstantArticlesClientEditor(props) {
+  /* Renders a panel with fields to edit the Facebook client information */
+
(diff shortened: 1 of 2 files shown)


dispatch/static/manager/src/js/components/integrations/FBInstantArticles.js
```diff
@@ -0,0 +1,141 @@
...
+
+function InstantArticlesPageDropdown(props) {
+  /* Renders a dropdown list to select a Facebook page */
+
+  const pageMap = props.pages.reduce((pageMap, page) => {
+    pageMap[page.id] = page
+    return pageMap
+  }, {})
+
+  const options = props.pages.map(page => {
```
JimFung: probably could have used an implicit return here.

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
```diff
@@ -264,6 +176,9 @@ const mapDispatchToProps = (dispatch) => {
     },
     updateIntegration: (integrationId, data) => {
       dispatch(integrationActions.updateIntegration(integrationId, data))
+    },
+    resetURI: (pathname) => {
```
JimFung, line 180: you can make this 1 line case it works without the {} around the dispatch call afaik. you can leave it in if you think it makes it look cleaner/more readable.

JimFung:
the rest LGTM, feel free to merge after changes 👍

dispatch/static/manager/src/js/pages/Integrations/FBInstantArticlesIntegrationPage.js
psiemens: yeah i think i prefer it on a separate line

psiemens:
Thanks @JimFung!

PR #493, opened 2017-04-05
Change Itemlist to take createHandler instead of createRoute

JamieRL:
Changed to pass function to return jsx for ArticleIndex.


psiemens:
Looks great 👍 

Feel free to merge

PR #494, opened 2017-04-06
475 file upload page

JamieRL:
Implement files page, and files-api tests. Minor change in Integration Serializer


dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -0,0 +1,78 @@
...
+import * as types from '../constants/ActionTypes'
+import { fileSchema } from '../constants/Schemas'
+import DispatchAPI from '../api/dispatch'
+
+
+export function fetchFiles(token, query) {
+  return {
+    type: types.FETCH_FILES,
+    payload: DispatchAPI.files.fetchFiles(token, query)
+      .then(function(json) {
```
JimFung: you should use the arrow syntax here to keep in style with how this is written in the other actions.

dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -0,0 +1,78 @@
...
+        }
+      })
+  }
+}
+
+export function createFile(token, file) {
+  return {
+    type: types.CREATE_FILE,
+    payload: DispatchAPI.files.createFile(token, file)
+      .then(function(json) {
```
JimFung: same as above, this should use the arrow syntax

dispatch/static/manager/src/js/pages/FilesPage.js
```diff
@@ -1,7 +1,176 @@
...
+    return prevProps.location.query.page !== props.location.query.page
+  }
+
+  handleDeleteFiles(fileIds) {
+    this.props.deleteFiles(this.props.token, fileIds)
+    this.props.clearSelectedFiles()
+  }
+
+  handleSearchFiles(query) {
+    this.props.searchFiles(this.props.token,query)
```
JimFung: need a space after the comma

JimFung (commented):
Mostly stylistic changes. good job 👍

JimFung:
LGTM on the js side, probably need @psiemens to take a look at the python

dispatch/apps/api/views.py
```diff
@@ -407,6 +407,42 @@ def articles(self, request, pk=None):
...
+        queryset = File.objects.all()
+        q = self.request.query_params.get('q', None)
+        if q is not None:
+            # If a search term (q) is present, filter queryset by term against `name`
+            queryset = queryset.filter(name__icontains=q)
+        return queryset
+
+    def bulk_delete(self, request):
+        deleted = []
+        ids = self.request.data.get('ids', None)
```
blx, line 427: `None` is the default value returned for `dict.get(nonexistent-key)`. It's generally more idiomatic to omit it:
```py
ids = self.request.data.get('ids')
```

This whole statement can be simplified into something like
```py
ids = self.request.data.get('ids', '').split(',')
File.objects.filter(id__in=ids).delete()

data = {
    'deleted': map(int, ids)
}
```

Does the frontend actually use the list of deleted IDs? What would be a case where only some deletions would fail?

dispatch/apps/api/views.py
```diff
@@ -407,6 +407,42 @@ def articles(self, request, pk=None):
...
+
+        if ids is not None:
+            ids = ids.split(',')
```
blx, line 430: Also, why the need to split ids? Why is the frontend passing a comma-separated string? Would it not be far more sane to pass a json array of ints?

dispatch/tests/test_api_files.py
```diff
@@ -0,0 +1,93 @@
...
+
+class FileTests(DispatchAPITestCase):
+
+    def _upload_file(self):
+        """
+        Upload a test file... somehow
+        """
+        url = reverse('api-files-list')
+
+        with open(os.path.join(os.path.dirname(__file__),'input/TestFile.txt')) as fp:
```
blx: You can also remove the need for an external file by using [`stringio`](https://docs.python.org/2/library/stringio.html), a builtin python library that creates file objects in memory:
```py
import StringIO
...

test_file = StringIO.StringIO('testtesttest')
data = {
    'name': 'TestFile',
    'file': test_file
}

...

test_file.close()

Comment (JamieRL) [reply dispatch/tests/test_api_files.py]: Oh, cool! That's exactly what I needed :)

File: dispatch/apps/api/views.py (hunk 5 above)
Comment (psiemens) [reply dispatch/apps/api/views.py:430]: This was copied from one of the other batch delete methods, which I think used to store the ids in the querystring (silly, I know). I agree with @blx that we should change it to a JSON array.

File: dispatch/apps/api/views.py (hunk 4 above)
Comment (psiemens) [reply dispatch/apps/api/views.py:427]: I think the motivation behind the separate queries was to catch errors on individual files (e.g. trying to delete a file that doesn't exist). However, this code doesn't actually have any proper error handling (not your fault, @JamieRL).

This method kinda falls outside of the standard REST architecture (not to mention the HTTP protocol), so I'm not sure what to do in the case where one or more of the files can't be deleted. Instinct tells me we should return some sort of error, but I'm unsure of the format. Do we return a 400 if only some of the operations fail? 

I think it makes the most sense to return an error if any of the operations fail; since the user requested to delete _all_ of the ids, it doesn't make sense to return 200 if that wasn't the case.

If we want to get really fancy we could use status code 207: http://www.restpatterns.org/HTTP_Status_Codes/207_-_Multi-Status

However, I'm leaning towards just scrapping the batch endpoint and refactoring the frontend to make multiple calls to the `article/{id}/delete` route. 

@blx @JamieRL thoughts?

File: dispatch/tests/test_api_files.py [hunk 7]
```diff
@@ -0,0 +1,93 @@
...
+from django.core.urlresolvers import reverse
+from rest_framework import status
+from dispatch.tests.cases import DispatchAPITestCase
+
+
+class FileTests(DispatchAPITestCase):
+
+    def _upload_file(self):
+        """
+        Upload a test file... somehow
```
psiemens: Haha can you change this to something more descriptive?

dispatch/tests/test_api_files.py
```diff
@@ -0,0 +1,93 @@
...
+                'file': fp
+            }
+
+            return self.client.post(url, data, format='multipart')
+
+    def _cleanup(self):
+        """
+        Delete created files and remove directory
+        """
+        shutil.rmtree(os.path.join(os.path.dirname(__file__),'media'))
```
psiemens: Add a space between arguments:

```py
os.path.dirname(__file__), 'media'
```

dispatch/tests/test_api_files.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # Check data
+        self.assertEqual(response.data['name'], 'TestFile')
+
+        self._cleanup()
+
+
+
+
```
psiemens: Extra whitespace here ^^

dispatch/tests/test_api_files.py
```diff
@@ -0,0 +1,93 @@
+import os
+import shutil
+import datetime
+from django.core.urlresolvers import reverse
+from rest_framework import status
+from dispatch.tests.cases import DispatchAPITestCase
+
```
psiemens: I like to split up my imports like so:

1. Python built-in libraries
2. External libraries
3. Django imports
4. Dispatch imports

```python
import os
import shutil
import datetime

from rest_framework import status

from django.core.urlresolvers import reverse

from dispatch.tests.cases import DispatchAPITestCase
```

dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -0,0 +1,76 @@
...
+
+export function clearFiles() {
+  return {
+    type: types.CLEAR_FILES
+  }
+}
+
+export function searchFiles(query) {
+  var queryObj = {}
+  if(query) {
```
psiemens: Add a space after `if`:

```js
if (query) {
```

dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -0,0 +1,76 @@
...
+    )
+  }
+}
+
+export function createFile(token, file) {
+  return {
+    type: types.CREATE_FILE,
+    payload: DispatchAPI.files.createFile(token, file)
+      .then( json => ({
+        result: json
```
psiemens: The result should be passed into normalizr:

```js
result: normalize(result, fileSchema)
```

dispatch/static/manager/src/js/pages/FilesPage.js
```diff
@@ -1,7 +1,176 @@
...
+    this.dropzone.open()
+  }
+
+
+  render() {
+    const title = 'Files'
+    const type = 'Files'
+    return (
+      <DocumentTitle title={title}>
+        <Dropzone ref={(node) => { this.dropzone = node }} className='c-files-dropzone' onDrop={(files) => this.onDrop(files)} disableClick={true} activeClassName='c-files-dropzone-active'>
```
psiemens: Put the properties on separate lines like this (it's just easier to read)

```jsx
<Dropzone 
    ref={(node) => { this.dropzone = node }} 
    className='c-files-dropzone' 
    onDrop={(files) => this.onDrop(files)} 
    disableClick={true}
    activeClassName='c-files-dropzone-active'>

File: dispatch/static/manager/src/js/api/dispatch.js [hunk 14]
```diff
@@ -30,8 +30,11 @@ function buildRoute(route, id) {
   return fullRoute
 }
 
-function buildHeaders(token) {
-  let headers = DEFAULT_HEADERS
+function buildHeaders(token, useDefaultHeaders=true) {
+  let headers = {}
+  if(useDefaultHeaders) {
+    headers = DEFAULT_HEADERS
+  }
```
psiemens: This works for now, but after you merge I'm going to modify this method to work better with a change we made in Jimmy's logout branch.

dispatch/tests/test_api_files.py
JamieRL: Oh lol, completely forgot I had put this there before I started

dispatch/static/manager/src/js/api/dispatch.js
JamieRL: Sounds good

JimFung: also, space after the if

dispatch/apps/api/views.py
blx: Most meaningful would probably be to remove the bulk endpoint, unless it's a real pain point for the frontend.

As it is currently, you're building a queryset (filter(IDs in `ids`)) and then calling delete on it -- that should never fail. If the frontend passes bad IDs, you'll just get an empty qs, which it's still valid to call delete on.

Apparently since django 1.3 deleting a model that has a FileField doesn't actually delete the file itself anyway, due to concerns about losing data in transaction rollbacks. ([django docs](https://docs.djangoproject.com/en/1.9/releases/1.3/#deleting-a-model-doesn-t-delete-associated-files))

There's some answers on [SO](http://stackoverflow.com/questions/16041232/django-delete-filefield) and a link to this apparently unmaintained [library](https://github.com/un1t/django-cleanup).

psiemens: > FileField doesn't actually delete the file itself anyway

This is an even trickier problem because Django allows you to specify custom storage handlers. For example, ubyssey.ca uses Amazon S3 via the `django-storages` package, which handles file deletion on its own (quite well, actually).

psiemens:
Merging! Thanks @JamieRL this is solid

PR #500, opened 2017-04-13
Add support for image uploads

### LINKED ISSUE #449: Add image upload capability

psiemens:
Close #449 

Upload images by dropping them into the image manager or via the default file dialog.


dispatch/apps/api/serializers.py
```diff
@@ -81,11 +86,18 @@ class Meta:
...
             'width',
-            'height'
+            'height',
+            'created_at',
+            'updated_at'
         )
 
+    def validate(self, data):
+        # Only accept ASCII filenames
+        if data.get('img') and not all(ord(c) < 128 for c in data.get('img').name):
```
blx: maybe factor this out, so no need for the comment?
```py
class ImageSerializer(...):
    ...
    def validate(self, data):
        if data.get('img') and not all_ascii(data['img'].name)
...

def all_ascii(s):
    return all(ord(c) < 128 for c in s)

File: dispatch/static/manager/src/js/reducers/EntitiesReducer.js [hunk 2]
```diff
@@ -78,6 +78,10 @@ export default function entitiesReducer(state = initialState, action) {
       images: R.merge(state.images, action.payload.entities.images),
       image: R.merge(state.image, action.payload.entities.images)
     })
+  case `${types.CREATE_IMAGE}_FULFILLED`:
+    return R.merge(state, {
+      images: R.merge(state.images, action.payload.entities.images)
```
blx, line 83: you can also lift the inner merge(s) into `mergeWith`, but maybe overly concise:
```js
return R.mergeWith(R.merge, state, {images: action.payload.entities.images})

PR #503, opened 2017-04-17
Add Pages test

Axel-Jacobsen:


dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+class PagesTest(DispatchAPITestCase):
+
+    def _create_page(self):
+        """
+        Create Dummy Page
+        """
+
+        # Create test person
+        person = Person.objects.create(full_tame='Test Person')
+        person.save()
```
psiemens: Since `person` is never used, you can remove these two lines

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
+from django.core.urlresolvers import reverse
+
+from rest_framework import status
+
+from dispatch.tests.cases import DispatchAPITestCase
+
+from dispatch.apps.content.models import Page, Person
```
psiemens: Don't need to import `Person`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+          "snippet": "This is a test snippet",
+          "content": [
+            {
+              "type": "paragraph",
+              "data": "This is some paragraph text"
+            }
+          ]
+        }
+
+        response = self.client.post(url, data, format='json')
```
psiemens: You can just call `self._create_page()` here

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+        page = self._create_page()
+
+        # Generate detail url
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        # Change the content
+        newdata = {
+          "title": "Test Page",
+          "slug": "test-page",
+          "snippet": "This is a test snippet",
```
psiemens: Since we're making a `patch` request, we can omit the "title", "slug", and "snippet" fields from `newdata`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+
+        # First make a page
+        page = self._create_page()
+
+        # Generate detail url
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        # Change the content
+        newdata = {
```
psiemens: Call this `new_data` (easier to read IMO)

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+        # Generate detail url
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        self.client.credentials()
+
+        # Change the fields
+        newdata = {
+          "title": "New Test Page",
+          "slug": "New test-page",
+          "snippet": "This is a new test snippet",
```
psiemens: Same as above -- don't need to include all fields

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,239 @@
...
+        # Generate detail url
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        # Successful deletion should return 204
+        response = self.client.delete(url, format='json')
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+        # Can't delete a page that has already been deleted
+        response = self.client.delete(url, format='json')
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
psiemens: The expected status code here is `status.HTTP_404_NOT_FOUND`

psiemens (changes requested):

dispatch/tests/test_api_pages.py
Axel-Jacobsen: Sounds good where is Person used? From looking at the class, it has info for admin. Is this to differentiate the reporters vs. programmers, and what privileges do each have?

dispatch/tests/test_api_pages.py
Axel-Jacobsen: Sounds good- what is "slug"? Snippet is the paragraph of info on the story below the title correct?

dispatch/tests/test_api_pages.py
psiemens: The `Person` model is used to store information about authors and photographers, as well any users in the system. Not every `Person` is a user, but those that are have a corresponding `User` model that can be accessed via `person.user`

dispatch/tests/test_api_pages.py
psiemens: The `slug` is what determines the url for article/page, and is unique to each instance.

For example, the slug for this article is `ubc-researchers-works-to-fight-fake-news`

screen shot 2017-04-16 at 10 33 57 pm

dispatch/tests/test_api_pages.py
Axel-Jacobsen: Thanks!

dispatch/tests/test_api_pages.py
Axel-Jacobsen: Thanks!

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,226 @@
...
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_create_empty_page(self):
+        """
+        Create page should fail with empty payload
+        """
+
+        url = reverse('api-pages-list')
+
+        self._create_page()
```
psiemens: This should be:
 ```py
response = self._create_page()
```

File: dispatch/tests/test_api_pages.py [hunk 9]
```diff
@@ -0,0 +1,226 @@
...
+        data = {
+            "title": "Test Page",
+        }
+
+        response = self.client.post(url, data, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        # Check Data
+        self.assertTrue('content_json' in response.data)
```
psiemens: Sorry, forgot to mention that the `content_json` field has been renamed to just `content`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,226 @@
...
+        Ensure we can make a page
+        """
+
+        response = self._create_page()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # Check Data
+        self.assertEqual(response.data['title'], 'Test Page')
+        self.assertEqual(response.data['slug'], 'test_page')
```
psiemens: should be `test-page` instead of `test_page`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,226 @@
...
+        # First make a page
+        page = self._create_page()
+
+        # Generate detail url
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        # Change the fields
+        new_data = {
+          "title": "New Test Page",
+          "slug": "New test-page",
```
psiemens: A slug can't contain spaces (not a valid URL). This should be changed to `'new-test-page'`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,226 @@
...
+              "data": "This is some paragraph text"
+            }
+          ]
+        }
+
+        response = self.client.patch(url, new_data, format='json')
+
+        # Check data
+        self.assertEqual(response.data['title'], "New Test Page")
+        self.assertEqual(response.data['slug'], "New test-page")
```
psiemens: Same here: `new-test-page` instead of `New test-page`

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,226 @@
...
+            }
+          ]
+        }
+
+        response = self.client.patch(url, new_data, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+        # Check Data
+        self.assertEqual(response.data['content'][0]['data'], 'This is some paragraph text')
```
psiemens: An unauthorized request actually won't return any data, so this line can be removed

psiemens (commented):
After looking at the test results, I noticed a few more bugs in your code. Are you able to get the tests running locally? If not, shoot me a Slack message and I'll walk ya through it :metal:

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,210 @@
+from django.core.urlresolvers import reverse
+
+from rest_framework import status
+
+from dispatch.tests.cases import DispatchAPITestCase
+
```
psiemens: Remove this space

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,210 @@
...
+    def _create_page(self):
+        """
+        Create Dummy Page
+        """
+
+        url = reverse('api-pages-list')
+
+        data = {
+          "title": "Test Page",
```
psiemens: Use single quotation marks instead of double. You can just do a search and replace for all double quotation marks

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,210 @@
...
+
+        url = reverse('api-pages-list')
+
+        response = self.client.post(url, None, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_create_incomplete_page(self):
+        """
+        Creat Page should fail with missing required fields
```
psiemens: Typo: "Create page should fail when missing required fields"

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # Check Data
+        self.assertEqual(response.data['title'], 'Test Page')
+        self.assertEqual(response.data['slug'], 'test-page')
+        self.assertEqual(response.data['snippet'], "This is a test snippet")
+        self.assertEqual(response.data['content'][0]['type'], 'paragraph')
+        self.assertEqual(response.data['content'][0]['data'], 'This is some paragraph text')
+
```
psiemens: I should have mentioned this earlier, but I was thinking it would be good to add an additional assertion to check that the page was actually saved to the database. Something like this:

```py
try:
  page = Page.objects.get(pk=response.data['id'])
except Page.DoesNotExist:
  self.fail("The page should exist in the database")

self.assertEqual(page.title, 'Test page')
self.assertEqual(page.slug, 'test-page')
self.assertEqual(page.snippet, 'This is a test snippet')
self.assertEqual(page.content[0]['type'], 'paragraph')
self.assertEqual(page.content[0]['data'], 'This is some paragraph text')
```

Let's also do this for the update tests below

dispatch/tests/test_api_pages.py
```diff
@@ -0,0 +1,210 @@
...
+        # Clear the client's credentials
+        self.client.credentials()
+
+        # Generate detail URL
+        url = reverse('api-pages-detail', args=[page.data['id']])
+
+        response = self.client.delete(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
```
psiemens, line 205: Add a check to make sure the page wasn't deleted from the database

psiemens (changes requested):

PR #506, opened 2017-04-18
Pages list

### LINKED ISSUE #466: Implement page listing

psiemens:
Close #466 

Pages backend unit tests are being done by @Axel-Jacobsen in #503

### DIFF
--- a/dispatch/static/manager/src/js/actions/PagesActions.js
+++ b/dispatch/static/manager/src/js/actions/PagesActions.js
@@ -1,0 +1,177 @@
+import R from 'ramda'
+import { normalize, arrayOf } from 'normalizr'
+import { push } from 'react-router-redux'
+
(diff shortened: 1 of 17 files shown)


dispatch/static/manager/src/js/actions/PagesActions.js
```diff
@@ -0,0 +1,177 @@
...
+    // Convert page contentState to JSON array
+    data.content = ContentStateHelper.toJSON(data._content)
+    delete data._content
+  }
+
+  // Set template_id
+  data.template_id = data.template
+  delete data.template
+
+  if (R.has('featured_image', data) && data.featured_image) {
```
blx: why not just
```js
if (data.featured_image)
```
? it's only one key level, not nested, so if it is missing or falsey you'll just get `undefined` and fail the if anyway.

PR #510, opened 2017-04-28
Create generic actions class

psiemens:
@blx @JimFung @JamieRL can you guys let me know what you think of this approach to adding some abstraction to the Redux action generators? I only changed `ArticlesActions`, but if you like this approach I want to update the rest of the models as well.

The idea is that `GenericActions` takes a list of action types, an api client, and a schema and generates all the basic CRUD actions. You can simply extend `GenericActions` (as in the case of `PublishableActions`) to define additional actions. I'm not a fan of over-abstraction, but I thought this was a nice way to clean up our action generators. Let me know you if you disagree!


dispatch/static/manager/src/js/reducers/ArticlesReducer.js
```diff
@@ -66,33 +68,33 @@ function articlesListReducer(state = initialState.list, action) {
...
-  case types.FETCH_ARTICLE + '_PENDING':
+  case pending(types.ARTICLES.GET):
     return R.merge(state, {
       isLoading: true
     })
-  case `${types.PUBLISH_ARTICLE}_FULFILLED`:
-  case `${types.UNPUBLISH_ARTICLE}_FULFILLED`:
-  case `${types.FETCH_ARTICLE}_FULFILLED`:
-  case `${types.SAVE_ARTICLE}_FULFILLED`:
+  case fulfilled(types.ARTICLES.GET):
```
blx: 👍  this is much nicer

dispatch/static/manager/src/js/actions/ArticlesActions.js
```diff
@@ -1,210 +1,59 @@
...
-
-  if (section) {
-    queryObj.section = section
-  }
-
-  return function(dispatch) {
-    dispatch(push({ pathname: '/articles/', query: queryObj }))
-  }
-}
+export default new ArticlesActions(
```
blx, line 55: Do you think it would be cleaner to export an instance of a named class expression instead?
```js
export default new class ArticlesActions extends PublishableActions {
  prepareData(data) {
    ...
  }
  search(section, query) {
    ...
  }
}(
  types.ARTICLES,
  DispatchAPI.articles,
  articleSchema
)
```

dispatch/static/manager/src/js/actions/ArticlesActions.js
```diff
@@ -1,210 +1,59 @@
...
 
-export function previewArticle(token, articleId, data) {
+    data.tag_ids = data.tags
 
-  data = preparePayload(data)
+    data.topic_id = data.topic
 
-  return function(dispatch) {
-    dispatch({ type: `${types.SAVE_ARTICLE}_PENDING` })
+    data.template_id = data.template
```
blx, line 27: off topic of PR, and i haven't contributed to this development 🙃 , so sorry for being obnoxious, but what is up with the funky key-reassignment in this `prepareData` function? could the python (de)serializers, say, be changed to do this name translation?

dispatch/static/manager/src/js/actions/GenericActions.js
```diff
@@ -0,0 +1,132 @@
...
+  create(token, id, data) {
+    return {
+      type: this.types.CREATE,
+      payload: this.api.create(token, this.dataAdapter(data))
+        .then(json => normalize(json, this.schema))
+    }
+  }
+
+  delete(token, id, next=null) {
+    return (dispatch) => {
```
blx: another nit, but half of the single-var arrow fns seem to have `(arg) =>` and the other half don't (`arg =>`). is there a pattern there?

dispatch/static/manager/src/js/actions/GenericActions.js
```diff
@@ -0,0 +1,132 @@
...
+  deleteMany(token, ids) {
+    return (dispatch) => {
+      dispatch({ type: `${this.types.DELETE_MANY}_PENDING` })
+
+      Promise.all(
+        ids.map(id => this.api.delete(token, id))
+      )
+      .then(() => {
+        dispatch({
+          type: `${this.types.DELETE_MANY}_FULFILLED`,
```
blx: use the new ReducerHelpers here and in PublishableActions?

blx (commented):
dis is nice, i think it's worth continuing. 🙏

left a few comments of varying degrees of usefulness, but yeah overall i agree this cleans up a chunk of boilerplate.

dispatch/static/manager/src/js/actions/ArticlesActions.js
psiemens: Yeah this bugs me too. DRF makes it tricky to update nested relationships and effectively treats foreign key fields as read-only. I have a backend fix in mind, but in the meantime I added the `_id` fields along with this frontend code as a workaround.

I also feel a little uneasy about changing the data type of an API field between reads/writes. For example, on a `GET` the `template` field returns a JSON object with all of the template data, but all that's needed to update it is an integer id. Is it a no-op to use `template` for both operations? I'm not sure :thinking:

dispatch/static/manager/src/js/actions/ArticlesActions.js
psiemens: Hmm.. gonna think about this one

blx: i appreciate that it looks like i just spewed out all the keywords i could think of, but it should be valid! primary reason for would be that it more clearly shows that the class itself is a one-off.

psiemens: lol i don't doubt its validity -- just its readability. i know its a common method for singleton export but for some reason its always bugged me

dispatch/static/manager/src/js/actions/GenericActions.js
psiemens: yeee

JamieRL:
Like the idea, take some getting used to but overall a cleaner approach.

dispatch/static/manager/src/js/actions/ArticlesActions.js
blx: Hmm, ok. In some projects i've worked on, most of the more involved update endpoints are just manually implemented, so they can unpack data in a way that is cleaner for the frontend.

Letting fields be either an id or a full object is a lil weird but if you sidestep that, I think it's a useful pattern. As long as all the parties can understand both `{template: 46}` and `{template: {id: 46, fieldA: "xyz", ...}}` (ie. there is no confusion that it is a reference, proxy, whatever, as opposed to a numeric field value), using the same name for both cleans things up, i think.

dispatch/static/manager/src/js/actions/ArticlesActions.js
```diff
@@ -1,210 +1,59 @@
...
-    type: types.SAVE_ARTICLE,
-    payload: DispatchAPI.articles.saveArticle(token, articleId, data)
-      .then( json => normalize(json, articleSchema) )
-  }
-}
-
-export function createArticle(token, data) {
-
-  data = preparePayload(data)
+  prepareData(data) {
```
JimFung, line 12: is both data.section_id and data.section required? or is only one of them required. if it is the later then i suggest creating a new object and add the require properties instead of remapping them as this will duplicate the values.

dispatch/static/manager/src/js/actions/ArticlesActions.js
```diff
@@ -1,210 +1,59 @@
...
-}
 
-export function toggleAllArticles(articleIds) {
-  return {
-    type: types.TOGGLE_ALL_ARTICLES,
-    ids: articleIds
-  }
-}
+  search(section, query) {
+    var queryObj = {}
```
JimFung: Should be a let. 
Doesn't the linter catch this?

dispatch/static/manager/src/js/constants/ActionTypes.js
```diff
@@ -19,18 +19,21 @@ export const DELETE_SECTIONS = 'DELETE_SECTIONS'
...
+  DELETE: 'DELETE_ARTICLE',
+  DELETE_MANY: 'DELETE_MANY_ARTICLE',
+  PUBLISH: 'PUBLISH_ARTICLE',
+  UNPUBLISH: 'UNPUBLISH_ARTICLE',
+  SET: 'SET_ARTICLE',
+  TOGGLE: 'TOGGLE_ARTICLE',
+  TOGGLE_ALL: 'TOGGLE_ALL_ARTICLE',
+  CLEAR_SELECTED: 'CLEAR_SELECTED_ARTICLE',
+  CLEAR_ALL: 'CLEAR_ALL_ARTICLE'
+}
```
JimFung: so much cleaner imo

JimFung (commented):
Overall, I think this is a lot cleaner 👍

dispatch/static/manager/src/js/actions/ArticlesActions.js
psiemens: ah, apparently it doesn't! nice catch

psiemens:
Thanks for the feedback! Going to move forward with this

dispatch/static/manager/src/js/actions/ArticlesActions.js
psiemens: although it shouldn't make a difference in this case because the nearest parent block is the function, which would be the scope of `var` anyways, right?

JimFung: Yep, Brought it up just so we're consistent.

psiemens:
So this turned into a monster PR... my apologies

I decided to make a small module (`util/redux`) to hold various redux utilities:

`ResourceActions` - a class that hooks up the basic CRUD operations for a resource
`Reducer` - a blueprint class for building reducers, for example:

```js
const initialState = { message: '' }

let messageReducer = new Reducer(initialState)

messageReducer.handle('UPDATE_MESSAGE', (state, action) => R.merge(state, { message: action.message }))

// Get the actual reducer function
export messageReducer.getReducer()
```

`buildManyResourceReducer` - generates a reducer with hooks for a resource listing
`buildSingleResourceReducer` - generates a reducer with hooks for a single resource

`pending`, `fulfilled`, `rejected` - helper functions for async actions

Let me know what you guys think of this approach! I'd like to document all of it in the wiki, but before that happens I want to make sure it's logical. I find that it not only reduces boilerplate but also makes the code easier to read/understand, mainly because those reducer switch statements could get pretty ugly.

dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -1,91 +1,29 @@
...
-      })
-    })
-    .catch(error => {
-      dispatch({
-        type: `${types.DELETE_FILES}_REJECTED`,
-        payload: error
-      })
-    })
+  search(query) {
+    var queryObj = {}
```
psiemens: Should be a `let`

dispatch/static/manager/src/js/actions/FilesActions.js
```diff
@@ -1,91 +1,29 @@
...
-}
+    if (query) {
+      queryObj.q = query
+    }
 
-export function toggleFile(fileId) {
-  return {
-    type: types.TOGGLE_FILE,
-    id: fileId
+    return function(dispatch) {
```
psiemens: Change to arrow func

dispatch/static/manager/src/js/actions/IntegrationActions.js
```diff
@@ -3,7 +3,7 @@ import DispatchAPI from '../api/dispatch'
 
 export function integrationCallback(token, integrationId, query) {
   return {
-    type: types.INTEGRATION_CALLBACK,
+    type: types.INTEGRATIONS.CALLBACK,
     payload: DispatchAPI.integrations.callback(token, integrationId, query)
       .then(function(json) {
```
psiemens: Arrow func

dispatch/static/manager/src/js/actions/IntegrationActions.js
```diff
@@ -16,8 +16,8 @@ export function integrationCallback(token, integrationId, query) {
 
 export function fetchIntegration(token, integrationId) {
   return {
-    type: types.FETCH_INTEGRATION,
-    payload: DispatchAPI.integrations.fetchIntegration(token, integrationId)
+    type: types.INTEGRATIONS.GET,
+    payload: DispatchAPI.integrations.get(token, integrationId)
       .then(function(json) {
```
psiemens: Arrow func

dispatch/static/manager/src/js/actions/IntegrationActions.js
```diff
@@ -29,8 +29,8 @@ export function fetchIntegration(token, integrationId) {
 
 export function saveIntegration(token, integrationId, data) {
   return {
-    type: types.SAVE_INTEGRATION,
-    payload: DispatchAPI.integrations.saveIntegration(token, integrationId, data)
+    type: types.INTEGRATIONS.SAVE,
+    payload: DispatchAPI.integrations.save(token, integrationId, data)
       .then(function(json) {
```
psiemens: Arrow func

dispatch/static/manager/src/js/actions/IntegrationActions.js
```diff
@@ -42,8 +42,8 @@ export function saveIntegration(token, integrationId, data) {
 
 export function deleteIntegration(token, integrationId) {
   return {
-    type: types.DELETE_INTEGRATION,
-    payload: DispatchAPI.integrations.deleteIntegration(token, integrationId)
+    type: types.INTEGRATIONS.DELETE,
+    payload: DispatchAPI.integrations.delete(token, integrationId)
       .then(function() {
```
psiemens: Arrow func

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
+import { normalize } from 'normalizr'
+
+import { ResourceActions } from '../util/redux'
+
+export default class PublishableActions extends ResourceActions {
+
+  publish(token, id, data) {
+    return (dispatch) => {
+
+      dispatch({ type: `${this.types.PUBLISH}_PENDING` })
```
psiemens: Use `pending` helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+
+      this.api.save(token, id, this.prepareData(data))
+        .then(() => this.api.publish(token, id))
+        .then(json => {
+          dispatch({
+            type: `${this.types.PUBLISH}_FULFILLED`,
```
psiemens: Use `fulfilled` helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+            payload: {
+              data: normalize(json, this.schema)
+            }
+          })
+        })
+        .catch(error => {
+          dispatch({
+            type: `${this.types.PUBLISH}_REJECTED`,
```
psiemens: Use `rejected` helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+        .then(json => ({
+          data: normalize(json, this.schema)
+        }))
+    }
+  }
+
+  preview(token, id, data) {
+    return (dispatch) => {
+
+      dispatch({ type: `${this.types.SAVE}_PENDING` })
```
psiemens: Use `pending` helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+
+      this.api.save(token, id, this.prepareData(data))
+        .then(json => {
+          dispatch({
+            type: `${this.types.SAVE}_FULFILLED`,
```
psiemens: Helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+            payload: {
+              data: normalize(json, this.schema)
+            }
+          })
+
+          window.open(json.url, `_dispatch-preview-${json.id}`)
+        })
+        .catch(error => {
+          dispatch({
+            type: `${this.types.SAVE}_REJECTED`,
```
psiemens: Helper

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -0,0 +1,66 @@
...
+      this.api.save(token, id, this.prepareData(data))
+        .then(json => {
+          dispatch({
+            type: `${this.types.SAVE}_FULFILLED`,
+            payload: {
+              data: normalize(json, this.schema)
+            }
+          })
+
+          window.open(json.url, `_dispatch-preview-${json.id}`)
```
psiemens: Replace `_dispatch-preview` with constant?

dispatch/static/manager/src/js/reducers/ToasterReducer.js
```diff
@@ -21,90 +22,90 @@ export default function toasterReducer(toaster = {}, action) {
...
+  //
+  // // Integrations
+  // case `${types.SAVE_INTEGRATION}_FULFILLED`:
+  //   return showToast('Integration updated')
+  // case `${types.DELETE_INTEGRATION}_FULFILLED`:
+  //   return showToast('Integration removed')
+  // case `${types.INTEGRATION_CALLBACK}_FULFILLED`:
+  //   return showToast('Authentication successful')
+  // case `${types.INTEGRATION_CALLBACK}_REJECTED`:
+  //   return showToast(action.payload.detail, Intent.DANGER)
```
psiemens: finish these cases

PR #518, opened 2017-05-09
Axel misc tests

Axel-Jacobsen:


dispatch/tests/test___init__.py
```diff
@@ -0,0 +1,8 @@
+import dispatch.__init__
```
psiemens: You can just do
```py
import dispatch
```

dispatch/tests/test___init__.py
```diff
@@ -0,0 +1,8 @@
...
+from django.test import TestCase
+
+class __init__Test(TestCase):
+
+    def test_get_version(self):
+
+        return dispatch.__init__.__version__
```
psiemens: You don't need to include `__init__` here -- Python will automatically resolve the version like so:

```py
dispatch.__version__
```

dispatch/tests/test_execute.py
```diff
@@ -0,0 +1,8 @@
+from dispatch.core.management import execute_from_command_line
+from django.test import TestCase
+
+class execute_from_command_lineTest(TestCase):
```
psiemens: In Python it's convention to camelcase all class names. We can probably shorten this as well:

```py
class CommandLineTest(TestCase):

PR #520, opened 2017-05-10
516 test api imagegallery

joncppl:


dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+        Upload an image that can be linked by galleries
+        """
+        url = reverse('api-images-list')
+
+        with open(self.get_input_file('test_image.png'), 'rb') as test_image:
+            response = self.client.post(url, { 'img': test_image }, format='multipart')
+
+        return response.data['id']
+
+    def _create_gallery_only(self, id, attachment):
```
psiemens: Maybe change `attachment` to `attachments` just to imply that there can be multiple

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+          "attachment_json": attachment
+        }
+
+        return self.client.post(url, data, format='json')
+
+    def _create_gallery(self, id):
+        """
+        Create a gallery instance for further testing
+        """
+        id1 = self._upload_image()
```
psiemens: Change to `image_1` just to be more readable

dispatch/tests/test_api_imagegallery.py
psiemens: Maybe change `attachment` to `attachments` just to imply that there can be multiple

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+    def test_imagegallery_all(self):
+        """
+        Test that listing all the galleries works
+        """
+        # insert 2 galleries
+        gallery1, img_id11, img_id12 = self._create_gallery(0)
+        gallery2, img_id21, img_id22 = self._create_gallery(1)
+
+        url = reverse('api-galleries-list')
+        response = self.client.get(url, None, format='json')
```
psiemens: I think you can just leave out the data argument here instead of passing `None`

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+
+        self.assertEqual(response.data['title'], 'Gallery Title 1')
+        self.assertEqual(response.data['images'][0]['caption'], 'test caption 1')
+        self.assertEqual(response.data['images'][0]['image']['id'], img_id1)
+        self.assertEqual(response.data['images'][1]['caption'], 'test caption 2')
+        self.assertEqual(response.data['images'][1]['image']['id'], img_id2)
+
+    def test_imagegallery_get_invalidd(self):
+        """
+        Test that getting an invalid id returnsappropriate error
```
psiemens: Typo: add space after "returns"

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        id = -1
+        url = reverse('api-galleries-detail', args=[id])
+        response = self.client.get(url, None, format='json')
```
psiemens: Omit the second argument

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        Create gallery should fail with unauthenticated request
+        """
+        # Clear authentication credentials
+        self.client.credentials()
+
+        url = reverse('api-galleries-list')
+        response = self.client.post(url, None, format='json')
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
```
psiemens: To make this test a bit more robust, you should send a valid payload containing gallery data and then check to verify that nothing was added to the database, like so:

```py
count = ImageGallery.objects.count()
self.assertEqual(count, 0)
```

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,370 @@
...
+        url = reverse('api-galleries-list')
+
+        data = {
+          "title": "Gallery Title"
+        }
+
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        id = response.data['id']
```
psiemens: Probably don't need to assign this to a variable

psiemens (changes requested):

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
+from django.core.urlresolvers import reverse
+
+from rest_framework import status
+
+from dispatch.tests.cases import DispatchAPITestCase, DispatchMediaTestMixin
+from dispatch.apps.content.models import ImageGallery
+
+class ImageGalleryTests(DispatchAPITestCase, DispatchMediaTestMixin):
+    def _upload_image(self):
```
psiemens: Add a line break before the first method in the class

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def test_imagegallery_create_unauthorized(self):
+        """Create gallery should fail with unauthenticated request"""
+
+        url = reverse('api-galleries-list')
+        image_1 = self._upload_image()
+
+        # Clear authentication credentials (after uploading the image)
+        self.client.credentials()
+        data = {
```
psiemens: Add line break

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
...
+              "image_id": -1
+            }
+          ]
+        }
+
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_imagegallery_create_no_image_ids(self):
+        """ that an appropriate error is returned when an image gallery is
```
psiemens: Looks like there's a typo at the start of this comment

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
...
+                    "caption": "test caption 1",
+                    "image_id": img_id1
+                },
+                {
+                    "caption": "test caption 2",
+                    "image_id": img_id2
+                }
+            ]
+        }
+        gallery2 = self.client.post(url, data, format='json')
```
psiemens: Add line break

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
...
+        gallery, img_id1, img_id2 = self._create_gallery(1)
+        id = gallery.data['id']
+
+        # Clear authentication credentials
+        self.client.credentials()
+
+        url = reverse('api-galleries-detail', args=[id])
+        response = self.client.delete(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
psiemens, line 327: I forgot to mention this earlier, but we should check to make sure the gallery wasn't deleted after this call is made. This can be done either by making another API GET request or by querying the database manually like so:

```py
try:
  ImageGallery.objects.get(pk=id)
except ImageGallery.DoesNotExist:
  self.fail('The gallery should not have been deleted')
```

dispatch/tests/test_api_imagegallery.py
```diff
@@ -0,0 +1,380 @@
...
+                }
+            ]
+        }
+        gallery2 = self.client.post(url, data, format='json')
+        self.assertEqual(gallery2.status_code, status.HTTP_201_CREATED)
+
+        id2 = gallery2.data['id']
+        url = reverse('api-galleries-detail', args=[id2])
+        response = self.client.get(url, format='json')
+        self.assertEqual(response.data['title'], 'New Gallery Title')
```
psiemens: Maybe split this lines up with some line breaks for readability

psiemens (changes requested):
Just a few small things:
- Suggested a few places for added line breaks to improve code readability
- Use single quotation marks instead of double (sorry I know it's nit-picky 😬 )
- Add a sanity check to the delete unauthorized test to make sure the gallery wasn't accidentally removed from the DB

PR #525, opened 2017-05-12
522 tag list create edit

joncppl:
Add a factory to make section, tag, topic pages and editors

- Based on the existing Section management page.
- Add Topic and Tag management pages that use it.
- Edit the topic and tag api, actions, reducers to be consistent with sections
- replace instead of push after create (smoother experience)
- will go to previous page if all items on page were deleted


dispatch/static/manager/src/js/actions/TagsActions.js
```diff
@@ -1,10 +1,40 @@
...
 import { tagSchema } from '../constants/Schemas'
+
 import DispatchAPI from '../api/dispatch'
 
 import { ResourceActions } from '../util/redux'
 
-export default new ResourceActions(
+class TagsActions extends ResourceActions {
+
+  listNav(token) {
```
psiemens: Tags and Topics don't need the listNav action

dispatch/static/manager/src/js/actions/TopicsActions.js
```diff
@@ -1,10 +1,40 @@
...
 import { topicSchema } from '../constants/Schemas'
+
 import DispatchAPI from '../api/dispatch'
 
 import { ResourceActions } from '../util/redux'
 
-export default new ResourceActions(
+class TopicsActions extends ResourceActions {
+
+  listNav(token) {
```
psiemens: Tags and Topics don't need the listNav action

dispatch/static/manager/src/js/components/ListItemEditorComponentFactory/index.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+import R from 'ramda'
+import DocumentTitle from 'react-document-title'
+
+import ListItemToolbar from './ListItemToolbar'
+
+const NEW_LISTITEM_ID = 'new'
+
+function ListItemEditorComponentFactory(formClass, classString, AFTER_DELETE) {
```
psiemens: As discussed, change this to a component instead of a function. The `formClass`, `classString`, and `AFTER_DELETE` attributes can become members of the component class:

```js
this.formClass
this.classString
this.AFTER_DELETE
```
You should then be able to extend `ListItemEditorComponent` and override the constructor to define those attributes.

dispatch/static/manager/src/js/util/redux/actions.js
```diff
@@ -1,6 +1,6 @@
 import R from 'ramda'
 import { normalize, arrayOf } from 'normalizr'
-import { push } from 'react-router-redux'
+import { push, replace } from 'react-router-redux'
```
psiemens: Is `push` still used anywhere in this file?

psiemens (changes requested):

dispatch/static/manager/src/js/util/redux/actions.js
joncppl: by delete,
It makes sense to use replace there too so I'll change it as well, then push won't need to be imported

psiemens (approved):

PR #538, opened 2017-05-16
442 page editor

joncppl:


dispatch/static/manager/src/js/components/PageEditor/PageContentEditor.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+
+// import QuillEditor from './QuillEditor'
```
psiemens: We can remove this line

dispatch/static/manager/src/js/pages/Pages/NewPagePage.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+import PageEditor from '../../components/PageEditor'
+
+export default function NewSectionPage(props) {
```
psiemens: Should be `NewPagePage` (sounds funny I know)

dispatch/static/manager/src/js/pages/Pages/PagePage.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+import PageEditor from '../../components/PageEditor'
+
+export default function NewSectionPage(props) {
```
psiemens: Should be `PagePage`

psiemens (changes requested):
A few notes:
- I think we could benefit from having an `Editor` folder that holds components shared between the `ArticleEditor` and `PageEditor`
- `PageHeadline` and `ArticleHeadline` can probably be combined into a generic component (`EditorTitle`?)
- The `published_at` field in the list view shows "Invalid date" for unpublished pages. Can we show something like "unpublished" instead?

psiemens (approved):

PR #553, opened 2017-05-25
User api endpoints

arashout:
I believe I finished the api endpoints for Users. 
Couple of things to note:

- I handled duplicate email errors in the serializers.py file, I'm not sure if this was the correct decision
- I didn't implement a method of attaching a person to a user manually. Since this is done by default on User creation
- I completed most of the api tests for Users as well


dispatch/apps/api/serializers.py
```diff
@@ -14,13 +14,46 @@ class UserSerializer(DispatchModelSerializer):
...
+            'email',
+            'person_id',
+            'password_a',
+            'password_b',
         )
 
+    def create(self, validated_data):
+
+        instance = User()
+        if User.objects.filter(email=validated_data['email']).exists():
```
psiemens: What happens if you try to call `instance.save()` with a duplicate email? Does it raise an error? I'm thinking it might be better to catch the error when it is raised instead of trying to validate the data beforehand

dispatch/tests/test_api_persons.py
```diff
@@ -161,6 +161,8 @@ def test_delete_person(self):
         response = self.client.delete(url, format='json')
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
 
+        #TODO: Check database for deletion
```
psiemens: Implement this check

dispatch/tests/test_api_users.py
```diff
@@ -0,0 +1,222 @@
+from dispatch.tests.cases import DispatchAPITestCase, DispatchMediaTestMixin
+from dispatch.apps.content.models import Person, User
+from django.core.urlresolvers import reverse
+from rest_framework import status
+from django.conf import settings
+from test_api_persons import PersonsTests
```
psiemens: Reorganize these imports like so:

```py
from rest_framework import status

from django.core.urlresolvers import reverse
from django.conf import settings

from dispatch.apps.content.models import Person, User
from dispatch.tests.cases import DispatchAPITestCase, DispatchMediaTestMixin
from dispatch.tests.test_api_persons import PersonsTests

File: dispatch/tests/test_api_users.py [hunk 4]
```diff
@@ -0,0 +1,222 @@
...
+        """Creating a user with duplicate emails should fail"""
+        response1 = self._create_user('test@gmail.com')
+        response2 = self._create_user('test@gmail.com')
+
+        self.assertEqual(response1.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response2.status_code, status.HTTP_409_CONFLICT)
+
+    def test_unauthorized_user_creation(self):
+        """Test unauthorized user creation"""
+        #NOTE: Can't use helper method here, credential clear will break _create_person
```
psiemens: Change to shorter comment:

```py
# Create person before clearing credentials
```

dispatch/tests/test_api_users.py
```diff
@@ -0,0 +1,222 @@
...
+        """A test case to ensure a variety of bad passwords are not succesful"""
+        person_id = self._create_person("Attached Person").data['id']
+        url = reverse('api-users-list')
+        data = {
+            'email' : 'testemail123@ubyssey.ca',
+            'person_id' : person_id,
+            'password_a': 'Matching',
+            'password_b': 'NotMatching'
+        }
+        # PASSWORD TESTS
```
psiemens: Remove this comment

dispatch/tests/test_api_users.py
```diff
@@ -0,0 +1,222 @@
...
+        UPDATED_EMAIL = 'updateTest@gmail.com'
+        UPDATED_PASSWORD = 'updatedPassword'
+
+        url = reverse('api-users-detail', args=[response.data['id']])
+        data = {
+            'email' : UPDATED_EMAIL,
+            'password_a' : UPDATED_PASSWORD,
+            'password_b' : UPDATED_PASSWORD
+        }
+        #TODO: Figure out how to test that password is updating
```
psiemens: You can try authenticating with that email/password combination like this:

```py
from django.contrib.auth import authenticate
user = authenticate(username='john', password='secret')
```

If the authentication fails then user will be None

psiemens (commented):
Looks good! Just suggested a few changes

dispatch/apps/api/serializers.py
arashout: I'm not quite sure how to go about this @psiemens. When we wanted to deal with the Person database error for cascading deletions we changed the PersonViewSet perform_destroy method, in this case should I override the UserViewSet perform_create method in [apps/api/views.py](views file)?

dispatch/tests/test_api_users.py
arashout: I'm not really sure why but both passwords are failing to authenticate, might need some help on this as well

psiemens (approved):

dispatch/apps/api/serializers.py
```diff
@@ -38,6 +28,54 @@ class Meta:
...
+        instance = User()
+        return self.update(instance, validated_data)
+
+    def update(self, instance, validated_data):
+
+        instance.email = validated_data.get('email', instance.email)
+        instance.person = validated_data.get('person', instance.person)
+
+        if validated_data.get('password_a'):
+            instance.set_password(validated_data.get('password_a'))
```
blx: this can likely be simplified to the faster
```py
    instance.set_password(validated_data['password_a'])
```
given that the existence of `password_a` has already been confirmed

dispatch/apps/api/serializers.py
```diff
@@ -38,6 +28,54 @@ class Meta:
...
         )
 
+class UserSerializer(DispatchModelSerializer):
+    """
+    Serializes the User model.
+    """
+
+    email = serializers.EmailField(
+        required=True,
+        validators=[UniqueValidator(queryset=User.objects.all())]
```
blx, line 38: is this required, on top of the [unique db constraint](https://github.com/ubyssey/dispatch/blob/develop/dispatch/apps/core/models.py#L26) on the User model itself?

psiemens: I too thought DRF was supposed to automatically validate unique fields, but we were getting a database `IntegrityError` until we explicitly defined the validator. 😕

PR #565, opened 2017-05-31
559 event manager

joncppl:
- implement event manager
- changed image from FK to ImageField
- event form
- datetime input


dispatch/static/manager/src/js/components/inputs/DateTimeInput.js
```diff
@@ -0,0 +1,37 @@
...
+        <AnchorButton className='c-datetimeinput-button'>
+          <span className='pt-icon-standard pt-icon-calendar' />
+        </AnchorButton>
+      </Popover>
+      <div className='c-datetimeinput-text'>
+        {textString}
+      </div>
+    </div>
+  )
+}
```
psiemens: I'd like to change the design of the date field to make it a bit more condensed. Can you use a [control group](http://blueprintjs.com/docs/#core/components/forms/control-group) to do something like this?

screen shot 2017-06-02 at 5 05 59 pm
screen shot 2017-06-02 at 5 05 49 pm

Clicking on the calendar icon or on the field directly would open up the popover for editing

dispatch/static/manager/src/js/components/inputs/DateTimeInput.js
```diff
@@ -0,0 +1,37 @@
...
+  const textString = props.value ?
+    humanizeDatetime(props.value)
+    : 'Select a date and time...'
+
+  const picker = (
+    <DateTimePicker
+      value={props.value || new Date()}
+      onChange={props.onChange}
+      timePickerProps={{ // so that state updates when either
+        // date or time are changed
```
psiemens: Probably don't need these comments

dispatch/static/manager/src/js/components/inputs/DateTimeInput.js
```diff
@@ -0,0 +1,37 @@
...
+        onChange: props.onChange
+      }} />
+  )
+
+  return (
+    <div>
```
psiemens: Add `className='c-input--datetime'`

dispatch/static/manager/src/js/components/inputs/ItemSelectInput.js
```diff
@@ -128,7 +129,7 @@ class ItemSelectInput extends React.Component {
           isSelected={false}
           text={item[this.props.attribute]}
           onClick={() => this.addValue(item.id)} />
-      ))
+      )) : []
```
psiemens: Instead of this you can add `results` to the defaultProps: https://github.com/ubyssey/dispatch/blob/3eb48b97ef296d098c7159f12e7d0965298d88b1/dispatch/static/manager/src/js/components/inputs/ItemSelectInput.js#L176-L178

dispatch/static/manager/src/js/components/EventEditor/EventForm.js
```diff
@@ -0,0 +1,170 @@
...
+            value={this.props.listItem.address || ''}
+            fill={true}
+            onChange={ e => this.props.update('address', e.target.value) } />
+        </FormInput>
+
+        <FormInput
+          label='Category'
+          padded={false}
+          error={this.props.errors.category}>
+          <SelectInput
```
psiemens, line 150: This field doesn't seem to be working properly. It doesn't show the right category when editing an existing event

dispatch/static/manager/src/js/components/EventEditor/EventForm.js
```diff
@@ -0,0 +1,170 @@
+import React from 'react'
+import Dropzone from 'react-dropzone'
+import { AnchorButton } from '@blueprintjs/core'
+
+import { FormInput, TextInput, TextAreaInput, DateTimeInput, SelectInput } from '../inputs'
+
+const CATEGORY_CHOICES = [
+    { value: '', label: 'Please Select' },
```
psiemens: Is there a way to show this label without making it one of the choice?

psiemens (changes requested):
In addition to my code comments, I also noticed that the page title isn't being displayed (both in the browser tab and on the page itself)

dispatch/static/manager/src/js/components/EventEditor/EventForm.js
joncppl: Any particular case?
It's all working my end.

psiemens: Weird, seems to be working now 👍

dispatch/static/manager/src/js/components/EventEditor/EventForm.js
```diff
@@ -5,7 +5,7 @@ import { AnchorButton } from '@blueprintjs/core'
 import { FormInput, TextInput, TextAreaInput, DateTimeInput, SelectInput } from '../inputs'
 
 const CATEGORY_CHOICES = [
-    { value: '', label: 'Please Select' },
+    // { value: '', label: 'Please Select' },
```
psiemens: Remove commented line

dispatch/static/manager/src/js/components/inputs/DateTimeInput.js
```diff
@@ -13,25 +13,26 @@ export default function DateTimeInput(props) {
...
       }} />
   )
 
   return (
-    <div>
+    <div className='c-input--datetime'>
       <Popover
         content={picker}
         popoverClassName='pt-popover-content-sizing'
         position={Position.RIGHT}>
```
psiemens: Super nit-picky, but can this open underneath instead of to the right? Just to be consistent with other dropdowns, and so that this would still work if it was used in the sidebar

psiemens (commented):
Just two small things -- otherwise good to push! Datepicker looks great :metal:

PR #569, opened 2017-06-01
#560 itemlist headers

joncppl:
- Allow configurable column headers are configured in itemlist. 
- configure all pages using itemlist to have appropriate headers.
- (events will need to be updated to have headers following merge)
- The Article and Page index pages weren't using the reusable itemlist component. Now they do.

### DIFF
--- a/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
+++ b/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import R from 'ramda'
+
+export default function ItemListColumnHeaders(props) {
+  if (!props.headers) {
+    return null
+  }
+
+  // blank first column for checkbox
+  const headers = R.insert(0, '', props.headers)
+
+  const items = headers.map((elem, i) => {
+    return (
+      <div key={i} className='c-item-list__item__cell'>
+        {elem}
+      </div>
+    )
+  })
+
+  return (
+    <li className='c-item-list__item c-item-list__column__header'>
+      {items}
+    </li>
+  )
+}
--- a/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
+++ b/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
@@ -1,17 +1,13 @@
 import React from 'react'
-import R from 'ramda'
 
 export default function ItemListColumnHeaders(props) {
   if (!props.headers) {
     return null
   }
 
-  // blank first column for checkbox
-  const headers = R.insert(0, '', props.headers)
-
-  const items = headers.map((elem, i) => {
+  const items = props.headers.map((elem, i) => {
     return (
-      <div key={i} className='c-item-list__item__cell'>
+      <div key={i+1} className='c-item-list__item__cell'>
         {elem}
       </div>
     )
@@ -19,6 +15,7 @@ export default function ItemListColumnHeaders(props) {
 
   return (
     <li className='c-item-list__item c-item-list__column-headers'>
+      <div key={0} className='c-item-list__item__cell' />
       {items}
     </li>
   )
--- a/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
+++ b/dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
@@ -18,7 +18,7 @@ export default function ItemListColumnHeaders(props) {
   })
 
   return (
-    <li className='c-item-list__item c-item-list__column__header'>
+    <li className='c-item-list__item c-item-list__column-headers'>
       {items}
     </li>
   )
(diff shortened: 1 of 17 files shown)


dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import R from 'ramda'
+
+export default function ItemListColumnHeaders(props) {
+  if (!props.headers) {
+    return null
+  }
+
+  // blank first column for checkbox
+  const headers = R.insert(0, '', props.headers)
```
psiemens: In this case i think it might be cleaner to just write:

```js
const headers = [''] + props.headers
```

dispatch/static/manager/src/js/pages/ItemIndexPage.js
```diff
@@ -83,8 +84,20 @@ export default class ListItemsPageComponent extends React.Component {
   }
 
   render() {
-    // Make the title start with uppercase letter
-    const titleString = this.props.typePlural.replace(/^\w/, m => m.toUpperCase())
+    const titleString = this.props.pageTitle
+      // Make the title start with uppercase letter
+      || this.props.typePlural.replace(/^\w/, m => m.toUpperCase())
```
psiemens: Not sure if I love this approach, just because it's hidden from whoever is using the component. I think I'd prefer to make the `pageTitle` prop required

dispatch/static/manager/src/js/pages/Articles/ArticleIndexPage.js
```diff
@@ -1,190 +1,72 @@
...
-  componentWillMount() {
-    // Fetch articles
-    this.props.clearArticles()
-    this.props.clearSelectedArticles()
-    this.props.listArticles(this.props.token, this.getQuery())
-  }
-
-  componentDidUpdate(prevProps) {
-
-    if (this.isNewSection(prevProps, this.props) || this.isNewQuery(prevProps, this.props)) {
```
psiemens, line 65: I *really* like the abstraction provided by `ItemIndexPage`, but in this case we lose the `isNewSection` check, which prevents the user from switching between different section tabs. Can we maybe add an optional prop to `ItemIndexPage` that allows us to extend this function?

psiemens (changes requested):
Awesome stuff -- definitely a much-needed enhancement. I pushed a commit with some style tweaks and added a few comments to the code

dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
joncppl: I believe `const headers = [''] + props.headers` would downcast the operands into strings before concatenation, so I don't think it'd work.

Instead I'll just have the first pseudoelement be hardcoded in jsx

dispatch/static/manager/src/js/components/ItemList/ItemListColumnHeaders.js
```diff
@@ -1,24 +1,21 @@
...
-      <div key={i} className='c-item-list__item__cell'>
+      <div key={i+1} className='c-item-list__item__cell'>
         {elem}
       </div>
     )
   })
 
   return (
     <li className='c-item-list__item c-item-list__column-headers'>
+      <div key={0} className='c-item-list__item__cell' />
```
psiemens: Nice -- I like this solution

dispatch/static/manager/src/js/pages/Articles/ArticleIndexPage.js
```diff
@@ -58,7 +58,15 @@ function ArticlePageComponent(props) {
         item => item.published_at ? humanizeDatetime(item.published_at) : 'Unpublished',
         item => item.latest_version + ' revisions'
       ]}
-
+      extraRelistCondition={(prevProps, props) => {
```
psiemens: I think it makes more sense to name this something like `shouldReload`

dispatch/static/manager/src/js/pages/Articles/ArticleIndexPage.js
```diff
@@ -58,7 +58,15 @@ function ArticlePageComponent(props) {
...
+        return prevProps.location.query.section !== props.location.query.section
+      }}
+      extraQueryHandler={(query, props) => {
```
psiemens: Rename to `queryHandler` -- don't think extra needs to be included

dispatch/static/manager/src/js/pages/ItemIndexPage.js
```diff
@@ -24,6 +25,10 @@ export default class ListItemsPageComponent extends React.Component {
       query.listItem = this.props.location.query.listItem
     }
 
+    if (typeof this.props.extraQueryHandler === 'function') {
```
psiemens: Instead of checking the type here, it would be better to add these props to `propTypes` and then just do `if (this.props.extraQueryHandler)`

Alternatively, you could use `defaultProps` and a stub function like so to avoid the conditional entirely:

```js

ListItemsPageComponent.defaultProps = {
  extraQueryHandler: (query) => query
}
```

psiemens (changes requested):
Just a few suggestions -- looks good otherwise

Issue #568, opened 2017-06-01
Separate panel for pending events


psiemens:
A separate panel/page should be added to the events section of the manager to provide an interface for approving pending events. This page should show a list of events that have `is_submission=True` by using the `pending=1` flag in the API call (see #562). After clicking on an event, the user should be taken to a page that displays the event information along with an "Approve" and "Disapprove" button. Clicking "Approve" should update the `is_submission` flag via an API call and then redirect the user back to the pending events page. Clicking the "Disapprove" button should delete the event and return the user to the pending events page.

joncppl:
I made the interface for this use a tiles for each event, all viewed on the same page
Each tile has approve/edit/disapprove buttons. (edit links to normal edit page)
approve updates is_submission, disapprove deletes. Either will cause the tile to animate/ fade-out.

This way a user can view many pending events at once, without having to navigate to different pages.
For example, it would make rejecting a bulk of spammy events very fast.

PR #572, opened 2017-06-08
#568 event auditor

joncppl:
Notes:

-used a bit of component state in EventAuditor/index.js, mostly to keep the count of is_submission events correct without fetching the list again / writing a custom reducer.


dispatch/apps/api/views.py
```diff
@@ -403,6 +404,10 @@ def get_queryset(self):
                 Q(description__icontains=q) |
                 Q(host__icontains=q)
             )
+        if pending:
+            queryset = queryset.filter(
+                Q(is_submission__exact=1)
```
psiemens: You don't need to use `Q` here since there's only one condition. Also, the default queryset should exclude submitted events:

```py
if pending:
    queryset = queryset.filter(is_submission=True)
else:
    queryset = queryset.filter(is_submission=False)
```

Can you add a test case for this as well?

dispatch/static/manager/src/js/components/Header.js
```diff
@@ -54,6 +54,9 @@ export default function Header(props) {
                 <li className='c-header__dropdown__link'>
                   <Link to='/events/'><span className='pt-icon-standard pt-icon-timeline-events' /> events</Link>
                 </li>
+                <li className='c-header__dropdown__link'>
+                  <Link to='/events/audit'><span className='pt-icon-standard pt-icon-endorsed' /> audit events</Link>
+                </li>
```
psiemens: Instead of adding a separate link in the navigation, I think it would be better to link to this page from the main events listing. Something like this:

screen shot 2017-06-13 at 9 16 06 am

Clicking on "5 pending events" would take the user to the approval page.

I used the Blueprint tag component with the following classNames:

```html
<div class="pt-tag pt-large pt-intent-warning">5 pending events</div>
```

dispatch/static/manager/src/js/components/inputs/LinkButton.js
```diff
@@ -5,7 +5,12 @@ import { Button } from '@blueprintjs/core'
 export default function LinkButton(props) {
   return (
     <Link
-      to={props.to}>
+      to={props.to}
+      onClick={e => {
+        if (props.disabled) {
+          e.preventDefault()
```
psiemens: Nice addition 👍

dispatch/static/manager/src/js/index.js
```diff
@@ -100,6 +100,7 @@ render((
           <Route path='events'>
             <IndexRoute component={Pages.Events.Index} />
             <Route path='new' component={Pages.Events.NewEvent} />
+            <Route path='audit' component={Pages.EventAudit} />
```
psiemens: The audit page should be grouped with the other events pages and referenced like so:

`component={Pages.Events.Audit}`

dispatch/static/manager/src/js/pages/EventAudit/EventAuditPagination.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default function EventAuditPagination(props) {
+  const { page, pages } = props
+
+  return (
+    <div className='c-event-audit-pagination'>
+      <LinkButton
+        to={getPath(props.location, page - 1)}
+        disabled={page <= 1}>Prev</LinkButton>
+      {' '}
```
psiemens: Use `margin` to add space between buttons rather than a hardcoded space

dispatch/static/manager/src/js/pages/EventAudit/EventAuditPagination.js
```diff
@@ -0,0 +1,35 @@
...
+function getPath(location, page) {
+  let query = R.clone(location.query)
+
+  query.page = page
+
+  return {
+    pathname: location.pathname,
+    query: query
+  }
+}
```
psiemens: Is this done anywhere else in the application? Does it make sense to move to `util/helpers.js`?

dispatch/static/manager/src/js/pages/EventAudit/EventCard.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
+
+import { AnchorButton, Tooltip } from '@blueprintjs/core'
+import { LinkButton } from '../../components/inputs'
+import { humanizeDatetime } from '../../util/helpers'
+
+function labelAndProp(label, prop) {
```
psiemens: Convert this to a functional React component, then call like so below:

`<EventLabel label='Description' value={this.props.event.description} />`

dispatch/static/manager/src/js/pages/EventAudit/EventCard.js
```diff
@@ -0,0 +1,92 @@
...
+
+        <div className='c-event-audit-card-button-border' />
+        <div className='c-event-audit-card-button-row'>
+
+          <Tooltip
+            className='c-event-audit-card-button-right'
+            transitionDuration={15}
+            content="Delete">
+            <AnchorButton
+              disabled={this.state.disableButtons}
```
psiemens: Can you explain the need for `this.state.disableButtons`?

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+  pending: 1,
+  limit: PER_PAGE
+}
+
+class EventAuditPage extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      count: 0
```
psiemens: How hard would it be to control this in the store? I know it works this way, but it breaks out of the redux architecture and seems kinda hard to follow. I might be missing something though

Ideally the events list that we keep in the store would stay up to date, and any time an event goes from `is_submission=True` to `is_submission=False`, that change would be reflected. Then you can simply count up all of the events with `is_submission=True` and use that as your count

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+  getQuery(props) {
+    props = props || this.props
+
+    const query = R.clone(PENDING_QUERY)
+
+    if (props.location.query.page) {
+      query.offset = (Number(props.location.query.page) - 1) * PER_PAGE
```
psiemens: use `parseInt(props.location.query.page, 10)` instead of `Number`

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  approve(id) {
+    const event = this.props.entities.events[id]
+    event.is_submission = 0
+    this.props.saveEvent(this.props.token, id, event)
+    this.checkPage()
+    this.setState({ count: this.state.count - 1 })
+    this.reloadListFlag = false
+  }
```
psiemens: I think it might be better to create a new action for this, and then just call `this.props.approveEvent(this.props.token, id)`

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+  disapprove(id) {
+    this.props.deleteEvent(this.props.token, id)
+    this.checkPage()
+    this.setState({ count: this.state.count - 1 })
+    this.reloadListFlag = false
+  }
```
psiemens: Same for disapproving events

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+  render() {
+    this.eventCount = 0
+    const events = this.props.events.ids.map((id, idx) => {
```
psiemens: I prefer to chain this logic together like so:

```js
const events = this.props.events.ids
  .map(id => this.props.entities.events[id])
  .map((event, i) => {
    // ...
  })

File: dispatch/static/manager/src/js/pages/EventAudit/index.js [hunk 14]
```diff
@@ -0,0 +1,181 @@
...
+      const event = this.props.entities.events[id]
+
+      if (!event.is_submission) {
+        return null
+      }
+      this.eventCount++
+
+      return (
+        <EventCard
+          key={idx}
```
psiemens: Use `i` instead of `idx` (for consistency with rest of application)

dispatch/static/manager/src/js/pages/EventAudit/index.js
```diff
@@ -0,0 +1,181 @@
...
+          </h2>
+          <div className='c-event-audit-subtitle'>
+            {this.state.count > PER_PAGE ?
+              <EventAuditPagination
+                page={Number(this.props.location.query.page || 1)}
+                pages={Math.ceil(this.state.count / PER_PAGE)}
+                location={this.props.location} />: null}
+            {this.state.count} event{this.state.count == 1 ? '' : 's'} pending approval
+          </div>
+          <ReactCSSTransitionGroup
```
psiemens: This transitions are nice, but I think it'd prefer to remove them for now since we don't use them anywhere else in the application

psiemens (changes requested):
This is really good overall. I think we might disagree on some of the code design choices, so feel free to argue your side.

I try to manage the applications state entirely from inside the store so that components deal with as little "business logic" as possible. There are times when that pattern should be broken, but I'm not sure this is one of them.

dispatch/static/manager/src/js/pages/EventAudit/EventCard.js
joncppl: just to disable them during the animation.
Since I'll be removing that it won't be necessary anymore

dispatch/apps/api/views.py
joncppl: looks like axel has a test in Event-forms-etc. 
When that's merged I'll pull it into this branch.

Axel-Jacobsen: I completed this in my PR because we were both modifying the same thing. Feel free to modify as needed!

psiemens: Sweet! my bad -- didn't realize at first

psiemens (approved):

PR #573, opened 2017-06-08
Event forms etc

Axel-Jacobsen:


dispatch/apps/api/fields.py
```diff
@@ -11,13 +11,15 @@ def to_internal_value(self, data):
...
 
+
 class PrimaryKeyField(Field):
 
     def __init__(self, serializer, *args, **kwargs):
         super(PrimaryKeyField, self).__init__(*args, **kwargs)
 
         self.serializer = serializer
 
+
```
psiemens: Don't think we need these empty lines

dispatch/apps/api/views.py
```diff
@@ -411,6 +424,12 @@ def get_queryset(self):
                 Q(description__icontains=q) |
                 Q(host__icontains=q)
             )
+
+        elif pending is not None:
+            queryset = queryset.filter(
+                Q(is_submission__exact=True)
+            )
```
psiemens: Sync up with @joncppl on this, I left a comment on his PR

dispatch/tests/settings.py
```diff
@@ -13,6 +13,10 @@
 SECRET_KEY = 'fake-key'
 BASE_URL = 'http://test.dispatch/'
 
+FACEBOOK_CLIENT_ID = '910205815715611'
+FACEBOOK_CLIENT_SECRET = '30e3d954b915a87eb51524a29738e6e8'
```
psiemens: Change these to empty strings

dispatch/apps/events/facebook.py
```diff
@@ -0,0 +1,92 @@
...
+from django.conf import settings
+
+from dispatch.tests.apis import FacebookTest
+from dispatch.vendor.apis import Facebook, FacebookAPIError
+
+
+class FacebookEvent(object):
+
+    def __init__(self, facebook_url, facebook_class=Facebook):
+
```
psiemens: Change `facebook_class` to `api_provider`

dispatch/apps/events/forms.py
```diff
@@ -0,0 +1,9 @@
+from django.forms import ModelForm, DateTimeField
+from dispatch.apps.events.models import Event
+
+class EventForm(ModelForm):
+
+    class Meta:
+
+        model = Event
+        fields = ['title', 'description', 'host', 'image', 'start_time', 'end_time', 'location', 'address', 'category', 'facebook_url', 'facebook_image_url', 'is_submission', 'submitter_email', 'submitter_phone']
```
psiemens: Split these onto separate lines for readability

dispatch/apps/events/models.py
```diff
@@ -0,0 +1,61 @@
...
+    submitter_phone = CharField(max_length=15, validators=[phone_regex], null=True)
+
+    is_published = BooleanField(default=False)
+
+    def cacheimage(self):
+        """Store image locally if we have a facebook url"""
+
+        if self.facebook_image_url and not self.image:
+
+            result = urllib.urlretrieve(self.facebook_image_url)
```
psiemens: Can this call be made with the `requests` Python library instead of urllib?

dispatch/apps/events/models.py
```diff
@@ -0,0 +1,61 @@
...
+
+    facebook_image_url = TextField(max_length=500, null=True, blank=True)
+
+    is_submission = BooleanField(default=False, blank=True)
+
+    submitter_email = EmailField(null=True)
+
+    phone_regex = RegexValidator(regex=r'(\+?1-)?\(?([0-9]){3}(\)?)(|\s?|-?|\.?)([0-9]){3}(\)?|\s?|-?|\.?)([0-9]{4})$', message='Please enter a valid phone number')
+
+    submitter_phone = CharField(max_length=15, validators=[phone_regex], null=True)
```
psiemens: Is this worth using? https://github.com/stefanfoulis/django-phonenumber-field

psiemens (dismissed):
Only have time for a partial review right now -- will get back to this later today

dispatch/apps/events/models.py
Axel-Jacobsen: After reading through their documentation and playing around with the code I couldn't get it to work. Maybe I was implementing it incorrectly, but I was getting some weird responses. I think `urllib` will be easier

psiemens: Cool, works for me!

dispatch/apps/events/models.py
Axel-Jacobsen: I am going to branch off of Event-forms-etc and test it out! I'm thinking it would be worth it though to get rid of that regex string

dispatch/apps/events/views.py
```diff
@@ -0,0 +1,50 @@
...
+            if event.facebook_image_url and not event.image:
+                event.cacheimage()
+
+            event.save()
+
+            return redirect(success)
+
+        else:
+            print "ERRORS: " + str(form.errors)
+            return HttpResponse('Invalid event data: Try again')
```
psiemens: If you remove this `else` case, the form with render with the errors so that the user knows which fields are invalid. Right they are being taken to a different page that doesn't provide any feedback

dispatch/apps/events/facebook.py
```diff
@@ -0,0 +1,92 @@
...
+class FacebookEvent(object):
+
+    def __init__(self, facebook_url, api_provider=Facebook):
+
+        self.facebook_url = facebook_url
+        self.event_id = self.get_event_id(facebook_url)
+        self.api_provider = api_provider
+
+    def get_event_id(self, facebook_url):
+
```
psiemens: Add a simple docstring to explain what this method does

dispatch/apps/api/views.py
```diff
@@ -401,16 +403,33 @@ class EventViewSet(DispatchModelViewSet):
     serializer_class = EventSerializer
 
     def get_queryset(self):
-        queryset = Event.objects.all()
+        # Only authenticated users can see submission or unpublished events
+        if self.request.user.is_authenticated():
+            queryset = Event.objects.all()
+        else:
+            queryset = Event.objects.filter(
+                Q(is_submission__exact=False),
```
psiemens: I don't think you need to use `__exact` here -- `is_submission=False` should be enough

dispatch/apps/events/facebook.py
```diff
@@ -0,0 +1,92 @@
+import urllib2, requests, re, datetime
+
+from django.conf import settings
+
+from dispatch.tests.apis import FacebookTest
+from dispatch.vendor.apis import Facebook, FacebookAPIError
+
+
+class FacebookEvent(object):
+
```
psiemens: Add docstring to explain what this class does

dispatch/apps/events/facebook.py
```diff
@@ -0,0 +1,92 @@
...
+        if m:
+            return m.group(1)
+        else:
+            raise Exception('URL provided is not a valid facebook event url')
+
+    def get_json(self):
+        """Returns the json for the event linked by the facebook url"""
+
+        # Create instance of facebook
+        fb = self.api_provider()
```
psiemens: Perhaps `fb` should be initialized in the constructor since you're doing the same thing in `get_image`

Something like this:

```py

self.api = api_provider()
self.api.get_access_token({
    'client_id': settings.FACEBOOK_CLIENT_ID,
    'client_secret': settings.FACEBOOK_CLIENT_SECRET,
    'grant_type': 'client_credentials'
})

File: dispatch/apps/events/templates/submit.html [hunk 13]
```diff
@@ -0,0 +1,157 @@
...
+    {{ form.end_time.errors }}
+    <label for="{{ form.end_time.id_for_label }}">End Time:</label>
+    {{ form.end_time }}
+  </div>
+
+  <div class="fieldWrapper">
+    {{ form.location.errors }}
+    <label for="{{ form.location.id_for_label }}">Location:</label>
+    {{ form.location }}
+  </div>
```
psiemens, line 73: The date/time picker doesn't use the same datetime format as Django, so I get an error when I try to create an event. Is there a way to configure the picker to use the correct format, or perform some kind of conversion?

dispatch/apps/events/views.py
```diff
@@ -0,0 +1,50 @@
...
+from dispatch.apps.events.forms import EventForm
+from dispatch.apps.events.models import Event
+
+def success(request):
+
+    return HttpResponse('Your event has been created')
+
+def submit(request):
+
+    if request.POST.get("submit_facebook") and (request.POST.get('facebook_url') is not None):
```
psiemens: Use a variable to prevent calling `request.POST.get("facebook_url")` multiple times:

```py

facebook_url = request.POST.get('facebook_url', None)

if request.POST.get("submit_facebook") and facebook_url:
    ...

File: dispatch/apps/events/views.py [hunk 15]
```diff
@@ -0,0 +1,50 @@
...
+        # show form with facebook event data
+
+        event = FacebookEvent(request.POST.get('facebook_url'))
+
+        event_data = event.get_json()
+        event_data['facebook_url'] = request.POST.get('facebook_url')
+        event_data['facebook_image_url'] = event.get_image()
```
psiemens: I think it might make more sense to handle all of this inside the `FacebookEvent` class, and then call it like so:

```py
event_data = event.get_data()

form = EventForm(initial=event_data)

File: dispatch/tests/apis.py [hunk 16]
```diff
@@ -0,0 +1,103 @@
...
+class FacebookTest(Facebook):
+
+    API_ROOT = 'https://graph.facebook.com/v2.8'
+
+    def __init__(self, access_token=None):
+
+        self.access_token = access_token
+
+    def get_access_token(self, params):
+
```
psiemens: You can make this even simpler and just use an empty method:

```py
def get_access_token(self, params):
    pass
```

Same goes for the `__init__` method above

dispatch/tests/apis.py
```diff
@@ -0,0 +1,103 @@
+from dispatch.vendor.apis import Facebook, FacebookAPIError
+from dispatch.tests.cases import DispatchAPITestCase
+from dispatch.tests.helpers import DispatchTestHelpers
+
+class FacebookTest(Facebook):
```
psiemens: I don't think you have inherit from the original api class -- `object` is probably enough

psiemens (changes requested):

dispatch/apps/events/templates/submit.html
Axel-Jacobsen: Right I had this problem and had solved it, but not sure what steps I took. Will research and find out. What is your error message, and which lines are tripping up?

Axel-Jacobsen: What was the error message?

Axel-Jacobsen: try adding `'%Y/%m/%d %H:%M'` to the `'DATETIME_INPUT_FORMATS'` list in `formats.py`, path is `ubyssey-dev/lib/python2.7/site-packages/django/utils/formats.py`

joncppl: just letting you know that you could also modify `DATETIME_INPUT_FORMATS` in the local `settings.py` (eg. `dispatch/conf/default_settings.py`) so that you don't have to edit the django dist files

I think the datetimepicker takes a format argument as well. (might be more robust to configure it to return an ISO date string)
http://xdsoft.net/jqplugins/datetimepicker/#use_other_date_parser

psiemens (approved):

PR #581, opened 2017-06-14
Code embeds

JamieRL:
Added code embeds to content editor using react ace editor. Users can write custom HTML, CSS, and Javascript code and embed it in the article. Code is output with appropriate tags i.e HTML <div>...</div>, CSS <style>...<style>, JS <script>...<script>.  Does this makes sense for the css? I'm not sure <style>...<style> will mean anything outside <head>...</head> .


JamieRL:
Just realized I didn't write tests for the code embeds.....
Sorry been distracted with Japan stuff.

dispatch/apps/content/models.py
```diff
@@ -128,6 +128,14 @@ def prepare_html(nodes):
 
                     if node['type'] == 'paragraph':
                         html += "<p>%s</p>" % node['data']
+                    if node['type'] == 'code':
+                        if node.get('data',{}).get('mode',{}) == 'html':
```
joncppl: This can be cleaned up a bit.

- `node['data']` is already stored in `data` within the scope, so you can access it directly, your if statements would then be `if data['mode'] == 'html':`
- add a space after the comma which separates arguments, like so `.get('data', {})`

dispatch/static/manager/src/js/components/ContentEditor/embeds/CodeEmbed.js
```diff
@@ -0,0 +1,68 @@
...
+  render() {
+    return(
+      <div className='o-embed o-embed--code'>
+        <AceEditor
+          mode={this.props.data.mode}
+          theme='chrome'
+          width='100%'
+          height='300px'
+          value={this.props.data.embedValue}
+          defaultValue={this.props.data.embedValue ? this.props.data.embedValue : DEFAULTVALUE}
```
joncppl: this can simply be `this.props.data.embedValue || DEFAULT VALUE`

dispatch/static/manager/src/js/components/ContentEditor/embeds/CodeEmbed.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+
+import AceEditor from 'react-ace'
```
joncppl: react-ace is imported but hasn't been added to package.json or yarn.lock

rerun `yarn add react-ace` and commit the changes to those files

dispatch/static/manager/src/styles/components/embeds/code.scss
```diff
@@ -0,0 +1,23 @@
...
+  border-top-width: 1px;
+  border-top-color: $color-border-gray;
+  overflow: auto;
+}
+
+.o-embed--code_editor_mode_select {
+  float: right;
+}
+
+.o-embed--code {
```
joncppl: no need to include empty styles

dispatch/apps/content/models.py
```diff
@@ -128,6 +128,14 @@ def prepare_html(nodes):
...
+                            html += "<div>%s</div>" % node.get('data',{}).get('embedValue',{})
```
joncppl: `.get('embedValue', {})` should probably have a default return that is a string, such as `.get('embedValue', '')`

the same goes for `.get('mode')` above. No default is necessary since it'll return None and fail the conditional anyway.

dispatch/static/manager/src/js/components/ContentEditor/embeds/CodeEmbed.js
```diff
@@ -0,0 +1,68 @@
...
+import { Button } from '@blueprintjs/core'
```
joncppl: Button is imported but not used

joncppl (changes requested):
Looks good and works well.

As mentioned in the review comments, 'react-ace' is missing from package.json.
Everything else is just codestyle.

PR #583, opened 2017-06-14
577 confirm navigation

joncppl:
Fairly straightforward. Use react-router's api to confirm with the user whether they should navigate away from pages with editable content. Uses component state to track isSaved.


dispatch/static/manager/src/js/components/ArticleEditor/index.js
```diff
@@ -22,6 +23,8 @@ class ArticleEditorComponent extends React.Component {
     super(props)
 
     this.toggleStyle = this.toggleStyle.bind(this)
+
+    this.state = { isSaved: true }
```
psiemens: @joncppl Do you mind if I move the `isSaved` boolean to the store? I think it can be a default attribute for all of the `single` item stores, and should be pretty easy to update via the reducers.

joncppl: sure. I suppose that keeps it in check with the existing architecture.
I can make the change on Monday or you can do it.

PR #586, opened 2017-06-16
584 upcoming event widget

Axel-Jacobsen:


dispatch/tests/test_upcoming_event_widget.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertEqual(widget_data['events'][1]['title'], u'Title 3')
+        self.assertEqual(widget_data['events'][2]['title'], u'Title 4')
+        self.assertEqual(widget_data['events'][3]['title'], u'Title 5')
+
+    def test_widget_render(self):
+        """Test the rendering of the event widget"""
+
+        def days_in_future(days):
+            """returns datetime object `days` in the future"""
+            return (datetime.datetime.now() + datetime.timedelta(days=days)).strftime("%B %d, %Y, %I:%M %p").replace('AM', 'a.m.').replace('PM', 'p.m.').replace(" 0", " ")
```
Axel-Jacobsen: This line is convoluted, if you have a better solution please write it in!

dispatch/apps/events/models.py
```diff
@@ -47,6 +47,8 @@ class Event(Model):
 
     is_published = BooleanField(default=False)
 
+    slug = SlugField(max_length=255)
```
psiemens: Move this field below `title`

dispatch/theme/event_widget.py
```diff
@@ -0,0 +1,37 @@
...
+
+            # Pull the top five events, and automatically ignore the featured event
+            if event.id != self.data['featured_event']:
+
+                data['events'].append({
+                    'title': event.title,
+                    'date': event.start_time
+                })
+
+        return data
```
psiemens: You can simplify this method:

```py
data['events'] = Event.objects.filter(is_published=True).exclude(pk=data['featured_event']).order_by('start_time')[:5]

return data
```

The event objects can be passed directly to the template -- no need to pick out only the fields you want. This keeps the code simpler and adds for more flexibility if users want to create a custom template.

I didn't think of this earlier, but you'll also want to consider how to handle past events. I think it makes sense to show events that are happening *up to the current day*. So if it's 8pm and an event started at 7pm we'd still show it, but we wouldn't show it the next day. It'd be good to add a test case for this as well.

dispatch/theme/fields.py
```diff
@@ -43,10 +43,10 @@ class ModelField(Field):
     def validate(self, data):
         if self.many:
             if not isinstance(data, list) or not all([isinstance(id, int) for id in data]):
-                raise InvalidField('Data must be list of integers')
+                raise InvalidField('Data must be list of integers (i.e. the ids of the models)')
```
psiemens: Maybe just "Data must be a list of integer IDs"

dispatch/tests/test_upcoming_event_widget.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_no_featured_event(self):
+        """When the user forgets a featured event there should be 5 listed events"""
+
+        register.zone(Sidebar)
+        register.widget(UpcomingEventsWidget)
+        zone = Sidebar()
+        widget = UpcomingEventsWidget()
+
+        # Make 7 events with increasing start dates
+        for index in xrange(5):
```
psiemens: If you want to avoid writing `index+1`, you can use `xrange(1,6)`

dispatch/tests/test_widget_render.py
```diff
@@ -36,7 +36,7 @@ class TestEventWidget(Widget):
     description = TextField('Description')
     events = EventField('Featured Events')
 
-    def context(self,data):
+    def context(self, data):
 
         data['events'] = Event.objects.filter(is_published=True).order_by('-start_time')
```
psiemens: Change this to `data['events'] = Event.objects.filter(is_published=True)` for simplicity

dispatch/apps/events/models.py
psiemens: What happens if you just call `str(datetime.datetime.now() + datetime.timedelta(days=days))`?

psiemens (changes requested):
Looks good overall! Very thoroughly tested 👍

dispatch/apps/events/models.py
Axel-Jacobsen: You get something like `2017-06-20 12:19:13.111972` - the template language seems to format it differently

dispatch/tests/test_upcoming_event_widget.py
psiemens: What happens if you just call `str(datetime.datetime.now() + datetime.timedelta(days=days)))`?

psiemens (approved):

PR #590, opened 2017-06-20
Documentation

Axel-Jacobsen:
OK First draft of the git documentation and a little bit added to other files. I am fresh to git as well and I expect that I have made some errors, so please correct me where necessary. I would like to cut down the text by about half in basic_git.md, but I think this is a good starting point.


arashout (approved):
Thanks for fixing my typos and adding more content Axel!

PR #593, opened 2017-06-20
patch event view set to allow updating is_submission

joncppl:
also add test case

it changes the spec so that if an authenticated user lists events, they get pending=0&1, unless they specify pending.

makes it much easier to PATCH is_submission


Axel-Jacobsen (approved):
Looks good to me! Well done

PR #594, opened 2017-06-20
576 create topics tags from article page

joncppl:
- When ItemListInput is given a create prop, it puts an add button next to the query.
- create adds a new tag / topic with the existing action
- a new reducer grabs the newly created id
- itemlistinput detects that new id and immediately adds it


psiemens (commented):
This is sweet! It makes it so much easier to add tags/topics.

Instead of using the `newId` prop, would it be possible to create a new action `createTagArticle` that handles both creating **and** adding the tag to the article? If the article updates its tag list, the item list will update automatically. The current approach works, but I worry that it might introduce bugs if the `newId` prop isn't handled properly by future maintainers.

joncppl:
I had an implementation like that in https://github.com/ubyssey/dispatch/pull/594/commits/7162114b96c183cb5403d03204cdf1f8f3a913dd

but it felt messy. Is there a better way to make api calls sequentially?

psiemens:
I actually think I prefer your callback approach. It's breaking redux a little bit, but I think this case warrants it. I was picturing an action that both creates the tag and updates the article, but that would require a new action for every resource that uses tags. Your approach is definitely more flexible.

One slight modification: instead of creating a new `createAndAdd` action, could the default `create` method take an optional `callback` argument? That would make it easy to generalize this for all resources.

joncppl:
Good suggestion on the callback for the standard create action. Makes it much neater.

PR #595, opened 2017-06-20
confirm nav special case when creating

joncppl:
currently it asks the user if they want to navigate when creating a new item since the route goes from type/new to type/:id,

since isSaved doesn't get updated by the reducer in time, the confirm can be avoided by this simple regex check

### DIFF
--- a/dispatch/static/manager/src/js/components/ArticleEditor/index.js
+++ b/dispatch/static/manager/src/js/components/ArticleEditor/index.js
@@ -34,7 +34,7 @@ class ArticleEditorComponent extends React.Component {
     confirmNavigation(
       this.props.router,
       this.props.route,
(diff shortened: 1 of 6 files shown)


psiemens:
You're going to hate me for this, but I think I found a simpler solution.

When you dispatch the router update, you can include an optional `state` object:

```js
dispatch(replace({
  pathname: `${next}/${json.id}`,
  state: { ignoreLeaveHook: true }
}))
```

Then in the `leaveHook` you have access to the state object:

```js
(location) => {
  if (!location.state.ignoreLeaveHook && shouldConfirm()) {
    return 'Unsaved changes. Are you sure you want to leave?'
  }
  return null
}
```

joncppl:
That works well. I wish I had thought of it to save myself some time!

Issue #592, opened 2017-06-20
Fix getting into admin panel with back button


Axel-Jacobsen:
I can get access to the admin panel of http://localhost:8000/admin by following these steps:
1. Go to http://localhost:8000/admin (not http://localhost:8000/admin/login). It will redirect you to the login page
2. Type in the volunteer email and password to email and password fields
3. Click submit
4. Go back a page in your browser)

I then had access to the admin panel. I couldn't create any articles/events/etc.

Notes: 
- Only works with the volunteer credentials (i.e. if you try to use test@gmail.com and test for the email and password fields you don't get access)
- You can also get to any page in the admin page in a similar method: go to a page like http://localhost:8000/admin/events. It brings you to the login page, log in with the volunteer credentials, and then go back a page
- I think the issue is just redirecting to  http://localhost:8000/admin after validating the username/password combo is correct. We know the system is validating the user because when we go back the user has access to the admin panel.

joncppl:
Closing this because it's taken care of.

We can open a new issue for handling errors.

PR #596, opened 2017-06-21
Fixed the back button glitch into admin pannel, changed viewset to have more descriptive error message

Axel-Jacobsen:
All credit goes to Jon!
This takes care of #592


joncppl:
I like how using replace removes the login page from the browser history stack.

Perhaps the error message that the server returns should be fed back to the UI. As it is now there is no feedback as to whether credentials are wrong. vs. server is down, etc.

psiemens:
Sweet! Merging this now, but I agree with @joncppl that it would be useful to show display an error message in the future

PR #602, opened 2017-06-26
3 commits 3 bugfixes

joncppl:
1.  correct headline value
so that headline doesn't incorrectly retain value from previous edited article
2. datetimeinput minDate
rather than today, set minDate to BoT. This prevents raising an exception for events in the past.
also turn on  datepicker 'actions bar'
3. Editor handles new article case
create a new, blank contentstate in the event that one did not previously exist.
This was the simplest solution with the smallest change. Leaving content undefined caused various other errors.


dispatch/static/manager/src/js/vendor/dispatch-editor/components/Editor.js
```diff
@@ -88,13 +89,13 @@ class ContentEditor extends React.Component {
   componentWillReceiveProps(nextProps) {
     // Push content to editor state
     this.setState({
-      editorState: EditorState.push(this.state.editorState, nextProps.content)
+      editorState: EditorState.push(this.state.editorState, nextProps.content || ContentState.createFromText(''))
```
psiemens: Instead of passing in `ContentState.createFromText('')`, you should be able to wrap the whole call like this:

```js
if (nextProps.content) {
  this.setState({editorState: EditorState.push(this.state.editorState, nextProps.content)})
}
```

The component is initialized with an empty state, so there's no need to pass it in again. Give it a try, though -- I could be wrong.

dispatch/static/manager/src/js/vendor/dispatch-editor/components/Editor.js
```diff
@@ -88,13 +89,13 @@ class ContentEditor extends React.Component {
...
     })
   }
 
   componentWillMount() {
     // Push content to editor state
-    this.onChange(EditorState.push(this.state.editorState, this.props.content))
+    this.onChange(EditorState.push(this.state.editorState, this.props.content || ContentState.createFromText('')))
```
psiemens: Same goes for this one.

psiemens (changes requested):

dispatch/static/manager/src/js/vendor/dispatch-editor/components/Editor.js
joncppl: When I tried that, having undefined/null content state caused various other errors, so this was the smallest change that would seemingly fix everything.

In particular toJSON would need a special case.
Is that preferable?

joncppl:
I changed it around a bit, but not much has really changed.

I think having the editorState 'reset' on an undefined content is necessary.
otherwise if you save a new article then create another new article, the content from the first article is still there.

joncppl:
I also added to this PR a change to the way the store is created.
There are two versions, dev and production that are loaded in based on NODE_ENV.
This is the way I've seen it done in lots of boilerplates other projects.

That way we can have redux devtools & redux logger while developing, but they won't be bundled or turned on in production.

Otherwise nothing changed with it's function.

psiemens:
@joncppl Awesome. That will make development much nicer. It'd be great to add hot reloading at some point in the future.

psiemens (commented):
Is there a way to separate the dev/prod stores without having them in separate files? I'm just not a fan of all the duplicate code

joncppl:
The main reason behind separate files is to prevent development packages from being bundled. 
But ya, I can probably add a common file to reduce the amount of redundant code

joncppl:
now it's all in one neat file.

psiemens (approved):

Issue #589, opened 2017-06-19
Add publish button to event editor


psiemens:
Right now there is no way to publish events from Dispatch. A publish button should be added to the UI next to the save/delete buttons, and should also indicate the `is_published` status of the event.

Axel-Jacobsen:
Would I be able to merge the publish button to the style changes (#603 TO #608)? From there I'll style the publish button like the update/delete button.

PR #603, opened 2017-06-27
589 publish button

Axel-Jacobsen:
```
Publish button is working; however, I am getting a bug when I initially click on a new event from `localhost:8000/admin/events`. The first click doesn't redirect to the page, but raises a TypeError (`TypeError: undefined is not an object (evaluating 'props.entities.local[props.listItem.id].is_published'`) and changes the url to `localhost:8000/admin/events/<event id>`. Not sure of the solution, but will find one soon.
```
EDIT: This was solved in commit 157342f - Takes care of issue #589


joncppl:
When it's a new item, `props.listItem.id` is null. It throws an error because null is being used as an array index.

I recommend checking whether it's null first. If it *is* null don't render the publish button (or render a dummy, disabled button).

dispatch/static/manager/src/js/constants/ActionTypes.js
```diff
@@ -11,7 +11,7 @@ export const TOPICS = resourceActionTypes('TOPICS')
 export const TAGS = resourceActionTypes('TAGS')
 export const TEMPLATES = resourceActionTypes('TEMPLATES')
 export const GALLERIES = resourceActionTypes('GALLERIES')
-export const EVENTS = resourceActionTypes('EVENTS', ['COUNT_PENDING'])
+export const EVENTS = resourceActionTypes('EVENTS', ['COUNT_PENDING', 'PUBLISH', 'UNPUBLISH'])
```
joncppl: `PUBLISH` and `UNPUBLISH` action types aren't used, so they aren't necessary here.

dispatch/static/manager/src/js/reducers/ToasterReducer.js
```diff
@@ -184,6 +184,15 @@ export default function toasterReducer(toaster = {}, action) {
     return showToast(`${action.payload.length} event${action.payload.length > 1 ? 's' : ''} deleted`)
   case rejected(types.EVENTS.DELETE_MANY):
     return showToast('Some events could not be deleted', Intent.DANGER)
+  case fulfilled(types.EVENTS.PUBLISH):
```
joncppl: these publish actions are never fired off, and so these toast messages are never used.

joncppl (changes requested):
Great work on your first front-end/react issue.

A few things in addition to 
the line comments.

1. publish button is unstyled
You can use the button component from the blueprint framework to make it look consistent
http://blueprintjs.com/docs/#core/components/button

2. Error when you try to publish a new, unsaved event.
Might as well hide it or disable it in this case.

Axel-Jacobsen:
Did the suggestions from the review by @joncppl, and added a back button to audit page. In terms of the publish button styling, I wrote it as the "PRIMARY" style cause I thought blue would be good. I was also thinking that the success banner should be different? Same colour as publishing 
button, saying "Event Published" or "Event Unpublished". Thoughts?
screen shot 2017-06-29 at 2 49 27 pm

Issue #606, opened 2017-06-28
Total number of pages displayed as 0 when on audit page without events


Axel-Jacobsen:

joncppl:
Indeed. The pagination component shouldn't be shown when there are no items, as is the case with the ItemList page.

While I'm at it I'll add a back button as well to make it consistent with the editor pages.

PR #608, opened 2017-06-28
Audit page additions

Axel-Jacobsen:
Takes care of #606


dispatch/static/manager/src/js/pages/Events/EventAuditPage.js
```diff
@@ -75,6 +76,13 @@ class EventAuditPage extends React.Component {
     this.props.disapproveEvent(this.props.token, id)
   }
 
+  pagination() {
```
joncppl: I don't think this needs to be a function. A `const` in the render function should be sufficient.

dispatch/static/manager/src/js/pages/Events/EventAuditPage.js
```diff
@@ -93,19 +101,19 @@ class EventAuditPage extends React.Component {
...
+                <span className='pt-icon-standard pt-icon-arrow-left'></span>Back
+              </AnchorButton>
             {this.props.events.count} event{this.props.events.count == 1 ? '' : 's'} pending approval
             </ToolbarLeft>
             <ToolbarRight>
-              <ItemListPagination
-                currentPage={parseInt(this.props.location.query.page || 1, 10)}
-                totalPages={Math.ceil(this.props.events.count / PER_PAGE)}
-                location={this.props.location} />
+            {Math.ceil(this.props.events.count / PER_PAGE) ? this.pagination() : null}
```
joncppl: You can get away with 
`this.props.events.count ? pagination : null`

dispatch/static/manager/src/styles/components/event_audit.scss
```diff
@@ -73,3 +73,8 @@
     margin-right: 0.75em;
   }
 }
+
+.pt-button {
```
joncppl: Adding a style to `.pt-button` here will affect all pt-button in the application.
I'm not sure that is what you intended.
Is there a way to just select the one you added?

I think this change in https://github.com/ubyssey/dispatch/commit/8337541fdbd550bc05908ed5aa1f13e960e8b4d9
should cover it, so try merging that in instead.

joncppl (changes requested):
Good work.

Just a few style/organisational things

dispatch/static/manager/src/js/components/EventEditor/index.js
```diff
@@ -59,17 +61,50 @@ const mapDispatchToProps = (dispatch) => {
...
+  const unpublishButton = (
+    <AnchorButton onClick={() => props.unpublishEvent(props.token, props.listItem.id)}
+      intent={Intent.PRIMARY}
+      disabled={props.isNew} >
+      <span className='pt-icon-standard pt-icon-th'></span>
+      Unpublish
+    </AnchorButton>
+  )
+
+  const event_is_published = props.entities.local && props.listItem.id
```
psiemens: Use camelcase instead of underscores: `eventIsPublished`

psiemens: Also, this could probably be moved into a helper outside of the component and called like this:

```js
extraButton={isPublished(props) ? unpublishButton : publishButton}
```

psiemens (changes requested):
This is great! Super impressive for your first frontend JS task. Just a few small changes 🤘

dispatch/static/manager/src/js/components/EventEditor/index.js
```diff
@@ -95,20 +95,25 @@ function EventEditorComponent(props) {
     </AnchorButton>
   )
 
-  const event_is_published = props.entities.local && props.listItem.id
-  ? props.entities.local[props.listItem.id].is_published : false // default if not rendered to 'false', i.e. not published
+  // const isPublished = props.entities.local && props.listItem.id
+  // ? props.entities.local[props.listItem.id].is_published : false // default if not rendered to 'false', i.e. not published
```
psiemens: Remove these comments

psiemens (changes requested):

psiemens (approved):

Issue #600, opened 2017-06-23
Delete buttons in dispatch should ask for confirmation


joncppl:
This had crossed my mind once or twice, and then Axel brought it to my attention again yesterday.

All across dispatch delete buttons work with a single click, deleting entries with no confirmation.

I suggest a DeleteButton component, that takes the delete method as a prop, but instead of running it onclick, it should open a popover, modal, or something to that affect, asking for confirmation of the delete.

The popover would have a second button which would actually run the delete action, and it would not intersect the button that opened the popover.

Deletes would require two clicks, making accidental deletion more unlikely, and giving users a moment of pause to ensure they really do want to delete it.

psiemens:
I really like the idea of a `DeleteButton` that has confirmation built-in. I agree that it's too easy to delete things right now.

A quick solution would be to use [window.confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm), which is what `react-router` uses to confirm navigation.

joncppl:
I've implemented it with http://blueprintjs.com/docs/#core/components/alert

looks nice and has simple implementation.

PR #609, opened 2017-06-29
#600 confirm delete

joncppl:
Implemented with [blueprint Alert](http://blueprintjs.com/docs/#core/components/alert)

Also I used component state to track the alert being open. I believe that this is acceptable usage of state.
My reasoning as it would add unnecessary complexity. In addition, if multiple ConfirmButtons ever get rendered at once, having the value in store would cause a major headache, so in this case it makes sense for it to be local to the component. 

Let me know if my reasoning is incorrect or if there is a counter argument. I'm still trying to get the hang of when to use local state vs. global store, hence my verbosity.

### DIFF
--- a/dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
+++ b/dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
@@ -19,9 +19,9 @@ export default class ConfirmButton extends React.Component {
 
   render() {
     return (
-      <div className='c-input-confirmbutton'>
+      <div className={`${this.props.className} c-input-confirmbutton`}>
         <AnchorButton
-          className={this.props.className}
+          className={this.props.buttonClassName}
           intent={this.props.intent}
           disabled={this.props.disabled}
           onClick={() => this.setState({ alertOpen: true })}>
--- a/dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
+++ b/dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
@@ -19,9 +19,9 @@ export default class ConfirmButton extends React.Component {
 
   render() {
     return (
-      <div className={`${this.props.className} c-input-confirmbutton`}>
+      <div className='c-input--confirm'>
         <AnchorButton
-          className={this.props.buttonClassName}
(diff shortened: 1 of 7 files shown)


dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
```diff
@@ -0,0 +1,44 @@
...
+
+  onConfirm() {
+    this.setState({alertOpen: false}, () => {
+      this.props.onConfirm()
+    })
+  }
+
+  render() {
+    return (
+      <div className={`${this.props.className} c-input-confirmbutton`}>
```
psiemens: Change classname to `c-input--confirm`

dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
```diff
@@ -0,0 +1,44 @@
...
+        <AnchorButton
+          className={this.props.buttonClassName}
```
psiemens: I'd probably change this to `this.props.className` and remove it from the div

dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
```diff
@@ -0,0 +1,44 @@
...
+        </AnchorButton>
+        <Alert
+          isOpen={this.state.alertOpen}
+          cancelButtonText='Cancel'
+          confirmButtonText='Delete'
+          iconName='trash'
+          onCancel={() => this.setState({ alertOpen: false })}
+          onConfirm={() => this.onConfirm()}
+          intent={Intent.DANGER}
+          >
```
psiemens: Move closing bracket up

dispatch/static/manager/src/js/components/inputs/ConfirmButton.js
```diff
@@ -0,0 +1,44 @@
...
+          <p>Are you sure you want to delete this?</p>
```
psiemens: This message should be a prop

psiemens (changes requested):
Looks good! Definitely a valid use for component state.

PR #618, opened 2017-07-14
widgets fixes

joncppl:
currently if zone has no widget attached to it, it throws an error. This makes it easy to break the site from the dispatch admin panel. 

this check just prevents a nonetype error and an empty string in that case, so the view can still render.

Also returning `None` causes the text None to actually be rendered. So I replaced the return with an empty string in the other try block.

### DIFF
--- a/dispatch/apps/frontend/templatetags/widgets.py
+++ b/dispatch/apps/frontend/templatetags/widgets.py
@@ -15,7 +15,7 @@ def zone(zone_id):
 
     try:
         return zone.widget.render()
-    except WidgetNotFound:
+    except (WidgetNotFound, AttributeError):
         pass
 
     return ''
--- a/dispatch/apps/frontend/templatetags/widgets.py
+++ b/dispatch/apps/frontend/templatetags/widgets.py
@@ -1,4 +1,3 @@
-
 from django import template
 
 from dispatch.theme import ThemeManager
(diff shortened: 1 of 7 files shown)


dispatch/apps/frontend/templatetags/widgets.py
```diff
@@ -11,6 +12,10 @@ def zone(zone_id):
     try:
         zone = ThemeManager.Zones.get(zone_id)
     except ZoneNotFound:
-        return None
+        return ''
+
+    widget = zone.widget
+    if widget is not None:
```
psiemens: I think I'd prefer this (I tend to avoid unnecessary variable assignment, but I'm not sure what's more readable):

```python
if zone.widget:
  return zone.widget.render()

**Comment by psiemens** [comment]
Nice catch 👍

File: dispatch/apps/frontend/templatetags/widgets.py [hunk 2]
```diff
@@ -11,6 +11,9 @@ def zone(zone_id):
...
         zone = ThemeManager.Zones.get(zone_id)
     except ZoneNotFound:
-        return None
+        return ''
 
-    return zone.widget.render()
+    if zone.widget:
+        return zone.widget.render()
+
+    return ''
```
psiemens: Can you add a quick test case for this?

psiemens (changes requested):

psiemens (approved):

PR #620, opened 2017-07-14
Fix broken EventSelectInput

psiemens:
This is _hopefully_ a more robust fix to the broken `EventSelectInput`. This is my bad -- I should have better communicated how widget field normalization works in order to prevent this.


joncppl:
Ah. That is indeed much better. Thanks!

PR #624, opened 2017-07-17
Widgets more field types

joncppl:
I wanted DateTime and Int fields for the upcoming events widget, so I added them.


dispatch/static/manager/src/js/components/fields/IntField.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+import { TextInput } from '../inputs'
+
+export default class IntField extends React.Component {
+  onChange(val) {
+    val = val.replace(/[^\d]/, '')
```
blx: as opposed to `/[^\d]/g`?

without the global flag you can't guarantee that all non-numbers will be removed, only the first character. is that intentional?

```js
'fx123abc'.replace(/[^\d]/, '')
// 'x123abc'

'fx123abc'.replace(/[^\d]+/, '')
// '123abc'

'fx123abc'.replace(/[^\d]/g, '')
// '123'

File: dispatch/tests/test_widget_fields.py [hunk 2]
```diff
@@ -1,7 +1,10 @@
+from datetime import datetime
+
 from dispatch.apps.content.models import Article, Image
 from dispatch.apps.events.models import Event
 from dispatch.theme import register
-from dispatch.theme.fields import CharField, TextField, ArticleField, ImageField, WidgetField, EventField, Field
+from dispatch.theme.fields import CharField, TextField, ArticleField, ImageField, \
```
blx, line 6: there is also the weakly-arguably slightly nicer parenthesyntax
```py
from dispatch.theme.fields import (
    CharField, TextField, ...
    WidgetField, EventField, ...
)

File: dispatch/tests/test_widget_fields.py [hunk 3]
```diff
@@ -691,3 +694,77 @@ def test_event_to_json_no_data(self):
...
+
+    def test_datetimefield_invalid_data(self):
+        testfield = DateTimeField('Test')
+        date_sz = 'garbage'
+
+        try:
+            testfield.validate(date_sz)
+            self.fail('Field date is invalid, exception should have been thrown')
+        except InvalidField:
+            pass
```
blx, line 718: there is also built-in support in django.TestCase via unittest for expecting an exception:
```py
with self.assertRaises(InvalidField):
    testfield.validate(date_sz)
```

dispatch/theme/fields.py
```diff
@@ -101,6 +103,44 @@ def validate(self, data):
...
+
+class UIntField(IntField):
+
+    type = 'uint'
+
+    def validate(self, data):
+        val = super(UIntField, self).validate(data)
+
+        if val < 0:
+            raise InvalidField('%s must be greater than or equal to  0' % self.label)
```
blx: extra space before 0?

dispatch/static/manager/src/js/components/fields/IntField.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+import { TextInput } from '../inputs'
+
+export default class IntField extends React.Component {
```
psiemens: I'd prefer to be little more verbose and call this `IntegerField`

dispatch/static/manager/src/js/components/fields/IntField.js
```diff
@@ -0,0 +1,23 @@
...
+    this.props.onChange(Math.floor(val))
+  }
+
+  render() {
+    return (
+      <TextInput
+        placeholder={this.props.label}
+        value={this.props.data || ''}
+        fill={true}
+        type="text"
```
psiemens: You should be able to use `type="number"`, which I think removes the need for the regex validation above.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

dispatch/tests/test_widget_fields.py
psiemens: Nice, didn't know about this

dispatch/theme/fields.py
```diff
@@ -101,6 +103,44 @@ def validate(self, data):
...
+            return int(data, base=10)
+        except ValueError:
+            raise InvalidField('%s must be integer' % self.label)
+
+    def prepare_data(self, data):
+        if isinstance(data, int):
+            return data
+        return int(data, base=10)
+
+class UIntField(IntField):
```
psiemens: Instead of using a separate field, could the `IntegerField` accept optional `min_value` and `max_value` keyword args? In my opinion, that would be a cleaner approach, and it's also the way Django configures its built-in db `IntegerField`: https://docs.djangoproject.com/en/1.11/ref/forms/fields/#integerfield

psiemens: You should be able to accomplish that be overriding the `Field` constructor

dispatch/static/manager/src/js/components/fields/IntField.js
joncppl: type="number" doesn't neatly support integers. 

ie. it allows decimal places, and doesn't call onChange when they are inserted

joncppl:
I've addressed most of the concerns.
I agree that a single configurable IntegerField is a better option, so I made that change.

Also noticed that field errors weren't being displayed. That was an easy fix.

dispatch/static/manager/src/js/components/fields/IntegerField.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  onChange(val) {
+    let { selectionStart, selectionEnd } = this.refs.input.refs.input
+    const initialLength = val.length
+
+    val = val.replace(/[^\d-]+/g, '')
+
+    // allow a '-' at start, but nowhere else
```
joncppl, line 21: As I am not a regex master, I could not find a way to implement this in a single line. Perhaps it is possible?

dispatch/static/manager/src/js/components/fields/IntegerField.js
```diff
@@ -0,0 +1,55 @@
...
+    const dLength = val.length - initialLength
+    selectionStart += dLength
+    selectionEnd += dLength
+
+    this.setState({ selectionStart, selectionEnd }, () => this.props.onChange(val))
+  }
+
+  componentDidUpdate() {
+    //put the caret back in the correct place
+    this.refs.input.refs.input.setSelectionRange(
```
joncppl, line 39: This just puts the text cursor back in the right place after using a regex to manipulate the contents. Otherwise it goes flying to the end when an invalid character is typed.

Wouldn't have to do all this if input's had a type="integer" or safari supported the pattern input attribute.

dispatch/static/manager/src/js/components/fields/IntegerField.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  render() {
+    return (
+      <TextInput
+        ref='input'
+        placeholder={this.props.label}
+        value={this.props.data || ''}
+        fill={true}
+        type="text"
```
psiemens, line 51: Did you try using `type="number"` instead of text?

joncppl: https://github.com/ubyssey/dispatch/pull/624#discussion_r128326556

I don't like how it still lets you type decimal points

psiemens: Ah my bad, missed that

psiemens: What about this? https://www.w3schools.com/tags/att_input_step.asp 

I've never used it before, but might be something worth looking into. I do like your current approach because it performs all the necessary validation before sending the data to the server.

joncppl: I tried step. It makes little arrows that appear at the right hand side of the input field, which in addition to the arrow keys, (de)increment the value by step. However floating point values can still be entered manually.

Perhaps I am being overly pedantic; users will probably know they can't enter a fraction of an event and it will get caught by the server's field validation anyway.

dispatch/static/manager/src/js/components/fields/IntegerField.js
blx: one option is something like this, which could be further reduced into one line if you really wanted.
```js
const prefix = val[0] === '-' ? '-' : ''
val = prefix + val.replace(/[^\d]/g, '')
```

Issue #615, opened 2017-07-12
Content Editor doesn't escape html-like entities


joncppl:
Possible good but may be seen as weird behaviour:
Type `&copy` or other escape strings into the content editor and update. It turns into a ©. 

Definitely bad: HTML markup isn't escaped

Entering `<h1>Header</h1>`, updating, and it will stored as an h1 element. 
It will render in published page as an h1 element but not in the content editor, where it is uneditable.
However update once more and it goes back to a paragraph.

Anything that looks like an html tag gets swallowed.
eg. if a writer wanted to add  <> around something for stylistic reasons or because of the content of their article, it disappears!

psiemens:
Yikes, this isn't good. Any ideas for how to fix it?

joncppl:
probably a carefully placed call to an escapehtml function as the article content is updated.
I'm not sure where as I'm not super familiar with it.

PR #627, opened 2017-07-17
Escape article text content

joncppl:
Addresses https://github.com/ubyssey/dispatch/issues/615

### DIFF
--- a/dispatch/static/manager/src/js/vendor/dispatch-editor/helpers/convertJSON.js
+++ b/dispatch/static/manager/src/js/vendor/dispatch-editor/helpers/convertJSON.js
@@ -92,11 +92,11 @@ function fromJSON(jsonBlocks) {
   )
 }
 
-function doEscape(obj) {
-  if (obj.type == 'paragraph') {
-    obj.data = escape(obj.data)
+function escapeHTML(block) {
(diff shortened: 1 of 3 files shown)


dispatch/static/manager/src/js/vendor/dispatch-editor/helpers/convertJSON.js
```diff
@@ -92,6 +92,13 @@ function fromJSON(jsonBlocks) {
   )
 }
 
+function doEscape(obj) {
```
psiemens: I'd prefer to call this `escapeHTML` instead

dispatch/static/manager/src/js/vendor/dispatch-editor/helpers/convertJSON.js
```diff
@@ -92,6 +92,13 @@ function fromJSON(jsonBlocks) {
...
+  if (obj.type == 'paragraph') {
```
psiemens: Change `obj` to `block`

psiemens (changes requested):
Nice! Haven't had a chance to test it, but it's nice that it was such a simple fix

PR #629, opened 2017-07-23
New logout

joncppl:


dispatch/apps/api/views.py
```diff
@@ -431,23 +431,32 @@ def get_queryset(self):
...
+            data = {
+                'token': unicode(token)
+            }
+
+            return Response(data, status=status.HTTP_202_ACCEPTED)
+        else:
+            raise BadCredentials()
+
+    def user_unauthenticate(self, request):
+        #Get token for current user
```
psiemens: Comment not needed

dispatch/apps/api/views.py
```diff
@@ -431,23 +431,32 @@ def get_queryset(self):
...
+        token = get_object_or_404(Token, user=request.user)
+
+        token.delete()
 
-        return Response(data, status=status.HTTP_202_ACCEPTED)
-    else:
-        raise BadCredentials()
+        return Response({})
```
psiemens: This should return `HTTP_204_NO_CONTENT`

dispatch/apps/api/views.py
```diff
@@ -431,23 +431,32 @@ def get_queryset(self):
         return queryset
 
 
-@api_view(['POST'])
 @permission_classes((AllowAny,))
-def user_authenticate(request):
+class AuthenticationViewSet(viewsets.GenericViewSet):
```
psiemens: Do you think it might make sense to call this `TokenViewSet`? I'm undecided. 

We can treat a token like any other REST resource and then implement the `create` and `destroy` methods, which in this case would be synonymous with login and logout.

Let me know what you think

dispatch/tests/test_api_auth.py
```diff
@@ -0,0 +1,111 @@
...
+        logout_response = self.client.delete(url, {}, format='json')
+
+        # Try to get the token, if it doesn't exist, return None.
+        user = User.objects.get(email='test@test.com')
+        try:
+            token = Token.objects.get(user_id=user.id)
+        except Token.DoesNotExist:
+            token = None
+
+        self.assertIsNone(token)
```
psiemens: I think it might be cleaner to use [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) here

dispatch/tests/test_api_auth.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_login_with_no_params(self):
+        """
+        login() -> login fails, 404
+        """
+        url = reverse('auth-token')
+
+        data = {}
```
psiemens: Don't need this variable

dispatch/tests/test_api_auth.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        response = self.client.post(url, data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_logout_when_no_token_exists(self):
+        """
+        logout should fail, 404
+        """
+        # setup
```
psiemens: Remove this commment

dispatch/static/manager/src/js/pages/LogoutPage.js
```diff
@@ -0,0 +1,37 @@
...
+  componentDidMount() {
+    this.props.logoutUser(this.props.token)
+  }
+
+  render() {
+    return (
+      <div>Logging out...</div>
+    )
+  }
+}
```
psiemens: Add empty line space here

psiemens (changes requested):

dispatch/apps/api/views.py
joncppl: I like the idea of making it more consistent with the other REST resources.

I don't know django-rest-framework though so I imagine it'd take me a few hours or so to learn it well enough to make those changes.

dispatch/static/manager/src/js/pages/LogoutPage.js
joncppl: you mean a `<br />` in the div?

joncppl:
I used the backend code & tests that Jimmy did a while ago. I'll clean it up as you requested.

How is the redux-stuff? 

I may have a noticed a bug wherein it sometimes let's the user into the app when there is no token. Axel and I have had to manually clear our cookies to make it show us the login page a few times each. It hasn't happened consistently and I can't manage to repeat it though.

edit: Ok I think it was that the token had expired or gotten removed from the database. I think the reducer should handle an 'invalid token' response and wipe the cookies, send the user back to the login page. I'll make a separate issue for that though.

dispatch/apps/api/views.py
joncppl: or not. I think I worked it out already... will update this PR shortly

dispatch/static/manager/src/js/pages/LogoutPage.js
joncppl: oh. nvm. didn't see in the lines that were shown by github.

psiemens:
> I think the reducer should handle an 'invalid token' response and wipe the cookies, send the user back to the login page. I'll make a separate issue for that though.

Ah, that makes sense. Sounds good 👍

dispatch/static/manager/src/js/constants/ActionTypes.js
```diff
@@ -16,7 +16,8 @@ export const EVENTS = resourceActionTypes('EVENTS', ['COUNT_PENDING'])
 // Authentication actions
 export const AUTH = actionTypes('AUTH', [
   'LOGIN_REQUIRED',
-  'GET_TOKEN'
+  'GET_TOKEN',
+  'LOGOUT'
```
psiemens: Maybe the actions should be `CREATE_TOKEN` and `DELETE_TOKEN` for consistency?

psiemens (approved):

Issue #628, opened 2017-07-20
Send confirmation email to user after submitting event


psiemens:
After submitting an event via the public form, the user should receive an automatic confirmation email:

```
Thanks for submitting your event to The Ubyssey! 

AMS Events Presents: The 33rd Annual Welcome Back BBQ

You will receive a confirmation email once your event has been approved by our staff. In the meantime, please don't hesitate to contact us if you have any questions.

The Ubyssey Events Team
events@ubyssey.ca
```

Then, after their event has been **approved and published**, they should received another email:

```
Your event has been approved!

You can view your event here: https://www.ubyssey.ca/events/126123

Thanks again for submitting!

The Ubyssey Events Team
events@ubyssey.ca
```

The email messages can be html templates stored in `templates/events/email/success.html` and `templates/events/email/confirm.html`. To render the mail template, use [render_to_string](https://docs.djangoproject.com/en/1.11/topics/templates/#django.template.loader.render_to_string).

Here's an example of how to send an email with Django:

https://github.com/ubyssey/ubyssey-dispatch-theme/blob/develop/ubyssey/views/advertise.py#L19-L25

Then in your `settings.py` file:
```
EMAIL_HOST = 'smtp.gmail.com'
EMAIL_PORT = 465
EMAIL_HOST_USER = 'noreply@ubyssey.ca'
EMAIL_HOST_PASSWORD = 'xxxx'
EMAIL_USE_SSL = True
```

EDIT: Include a disapproval email, including a separate landing for the reason (include drop-down with common problems like 'Spam').

PR #631, opened 2017-07-24
Confirmation email

Axel-Jacobsen:
Takes care of #628


dispatch/apps/events/models.py
```diff
@@ -55,3 +63,22 @@ def save_image_from_url(self, url):
...
 
         self.save()
+
+
+@receiver(post_update, sender=Event)
+def send_published_email(sender, instance, **kwargs):
+
+    if instance.is_published and not instance.is_published_email:
+
+        template = render_to_string('events/email/success.html', {'id': instance.id})
```
psiemens: Change `template` variable name to `body`

dispatch/apps/events/views.py
```diff
@@ -32,6 +35,16 @@ def submit_form(request):
...
 
+            template = render_to_string('events/email/confirm.html', form.cleaned_data)
+
+            send_mail(
+                    'Your event has been created!',
+                    template,
+                    settings.EMAIL_HOST_USER,
+                    [form.cleaned_data['submitter_email']],
+                    fail_silently=True,
+                )
```
psiemens: This just occurred to me: for consistency with the published email, do you think it would make sense to send this when the `post_create` signal is fired?

dispatch/apps/events/views.py
```diff
@@ -32,6 +35,16 @@ def submit_form(request):
             form.is_submission = True
             form.save()
 
+            template = render_to_string('events/email/confirm.html', form.cleaned_data)
```
psiemens: Change to `body`

psiemens (changes requested):

dispatch/apps/events/views.py
Axel-Jacobsen: So that would be when the event is approved (and not yet published)? I think that makes more sense, I can implement that

dispatch/apps/events/models.py
```diff
@@ -68,15 +70,33 @@ def save_image_from_url(self, url):
 
 
 @receiver(post_update, sender=Event)
+def send_approved_email(sender, instance, **kwargs):
+
+    if not instance.is_approved:
```
psiemens: We should add an attribute similar to `is_published_email` to avoid sending multiple emails.

psiemens (changes requested):

dispatch/apps/events/models.py
Axel-Jacobsen: `is_approved` takes care of this, as it is never set to false. I'll rename it to `is_approved_email`

dispatch/apps/events/models.py
```diff
@@ -57,3 +67,40 @@ def save_image_from_url(self, url):
         )
 
         self.save()
+
+
+@receiver(post_update, sender=Event)
+def send_approved_email(sender, instance, **kwargs):
+
+    if not instance.is_approved_email:
```
psiemens: Shouldn't the check include the `is_submission` field?

I think it's possible to update an event without approving it, which would trigger this function and send out an email. Unless I'm missing something?

dispatch/apps/events/views.py
```diff
@@ -1,5 +1,8 @@
 from django.core.files.uploadedfile import SimpleUploadedFile
+from django.core.mail import send_mail
+from django.conf import settings
 from django.shortcuts import render, redirect
+from django.template.loader import render_to_string
 from django.http import HttpResponse
```
psiemens: I think you can remove these imports ^^

dispatch/apps/events/models.py
Axel-Jacobsen: Ah yes I will add that

dispatch/apps/events/views.py
Axel-Jacobsen: Yep sounds good!

Axel-Jacobsen:
Travis-CI Is failing because the email templates are in `ubyssey-dispatch-theme/develop`, not the master branch `ubyssey-dispatch-theme/staging`

dispatch/apps/events/models.py
```diff
@@ -57,3 +60,37 @@ def save_image_from_url(self, url):
...
         self.save()
+
+@receiver(pre_save, sender=Event)
+def send_approved_email(sender, instance, **kwargs):
+    """Send an email to the submitter when the event is approved."""
+    if instance.is_submission and not instance.is_submission_email:
+        body = render_to_string('events/email/submitted.html', {'title': instance.title})
+
+        send_mail(
+            'Your event has been approved!',
```
psiemens: Change subject to "Your event has been submitted!"

dispatch/apps/events/views.py
```diff
@@ -29,8 +29,10 @@ def submit_form(request):
         form = EventForm(request.POST, request.FILES)
 
         if form.is_valid():
-            form.is_submission = True
-            form.save()
+
```
psiemens: Remove blank line

psiemens (changes requested):
Just two small changes, then feel free to merge!

PR #642, opened 2017-07-28
Ubc event integration

Axel-Jacobsen:
This PR is for the UBC Event integration. It consists of the class to handle a UBC Event link, generalizations in `views.py` and `models.py` to handle the UBC Links and for any future links we want to support in the future. Mixes well with https://github.com/ubyssey/ubyssey-dispatch-theme/pull/297


dispatch/apps/events/views.py
```diff
@@ -13,16 +13,39 @@ def submit_success(request):
...
-    facebook_error = False
+    event_url = request.POST.get('event_url')
+    url_error = False
+
+    if request.POST.get('url_import') and event_url is not None:
+
+        handlers = {
+            'calendar.events.ubc.ca': {
+                'url_handler': UBCEvent,
+                'url_handler_error': UBCEventError
```
psiemens: Instead of using separate errors, I would just define a generic `EventError` that both classes use.

dispatch/apps/events/views.py
```diff
@@ -13,16 +13,39 @@ def submit_success(request):
     return render(request, 'events/submit/success.html')
 
 def submit_form(request):
-    facebook_url = request.POST.get('facebook_url')
-    facebook_error = False
+    event_url = request.POST.get('event_url')
+    url_error = False
+
+    if request.POST.get('url_import') and event_url is not None:
+
```
psiemens: Remove this empty line

dispatch/tests/test_facebook_event.py
```diff
@@ -1,4 +1,4 @@
-from dispatch.apps.events.facebook import FacebookEvent, FacebookEventError
+from dispatch.apps.events.event_integrations import FacebookEvent, FacebookEventError
```
psiemens: I would change `event_integrations` to `sources`

dispatch/apps/events/event_integrations.py
```diff
@@ -92,3 +100,63 @@ def get_image(self):
...
+
+        formatting_data = unformatted_data.replace('<br />', '\n')
+        formatting_data = formatting_data.replace('&amp;', '&')
+        formatting_data = formatting_data.replace('<td class="fieldname">', '\n\n')
+
+        tag_stripper = self.TagStripper()
+        description = tag_stripper.strip_tags(formatting_data)
+
+        return {
+            'description': description
```
psiemens: Can you scrape more fields than just the description? 

I'd take a look at [BeautifulSoup](https://www.crummy.com/software/BeautifulSoup/) if you haven't already. It makes it super easy to scrape data from HTML pages.

For example, you should be able to do something like this to get the title:

```py
title = soup.find('h2').string

File: dispatch/apps/events/event_integrations.py [hunk 5]
```diff
@@ -92,3 +100,63 @@ def get_image(self):
             raise FacebookEventError(FACEBOOK_ERROR_MESSAGE)
 
         return image_url
+
+class UBCEvent(object):
+    """UBC Event class"""
```
psiemens: Make this a bit more descriptive -- something like "A class to scrape event information from UBC event webpages."

dispatch/apps/events/views.py
```diff
@@ -13,16 +13,39 @@ def submit_success(request):
...
+        handlers = {
+            'calendar.events.ubc.ca': {
+                'url_handler': UBCEvent,
+                'url_handler_error': UBCEventError
+            },
+            'facebook.com/events/': {
```
psiemens: I really like the approach of using URL prefixes to match handlers, but I think it might be even better to use the domain name itself.

Can you try using regex to parse the hostname from the URL? In this case, the two valid hostnames would be:

```py
'calendar.events.ubc.ca'
'facebook.com'
```

Then you can define a simple lookup table:

```py
sources = {
  'calendar.events.ubc.ca': UBCEvent,
  'facebook.com': FacebookEvent
}

hostname = get_host_from_url_with_regex(url)

if hostname in sources:
  handler = sources[hostname]
  # ...
else:
  url_error = True

**Review by psiemens — CHANGES_REQUESTED** [review]
This is great! Great job extending the Facebook import and making it generic.

Feel free to let me know if you disagree with any of my suggestions

File: dispatch/apps/events/event_integrations.py (hunk 4 above)
Comment (Axel-Jacobsen) [reply dispatch/apps/events/event_integrations.py]: Sweet BeautifulSoup is a pleasure to use, what steps should I take to install it in dispatch? Same with the phone_number field (i.e. adding it to the installed apps w/ a version?)

File: dispatch/apps/events/views.py (hunk 6 above)
Comment (Axel-Jacobsen) [reply dispatch/apps/events/views.py]: I like this! I am going to keep the actual data collection from the handler wrapped in a try-except to catch any errors that occur in that process

File: dispatch/apps/events/event_integrations.py (hunk 4 above)
Comment (jimmyhli) [reply dispatch/apps/events/event_integrations.py]: I used beautifulSoup for something else last week, it seems to have an original beautifulSoup and a more recent version of it named BeautifulSoup4 lol, they aren't that different tho

Comment (Axel-Jacobsen) [reply dispatch/apps/events/event_integrations.py]: Oh perfect thanks!

File: dispatch/apps/events/sources.py [hunk 7]
```diff
@@ -104,13 +104,29 @@ def get_data(self):
...
+        time_strings = map(str.strip, time_string.split('-'))
+
+        try:
+            start_time = datetime.strptime(time_strings[0], '%A, %B %d, %Y %I:%M %p')
+            end_time = datetime.strptime(time_strings[1], '%A, %B %d, %Y %I:%M %p')
+        except ValueError:
+            start_time = datetime.strptime(time_strings[0], '%A, %B %d, %Y %I:%M %p')
+            date = start_time.date()
+            time = datetime.strptime(time_strings[1], '%I:%M %p').time()
+            end_time = datetime.combine(date, time)
```
Axel-Jacobsen: I'd just like to explain this snippet. There are two forms that UBCevents Calendar uses for their start and end times: if it is a same day event, it has the form of `Saturday, August 12, 2017 9:00 AM - 1:00 PM`, and if it is multi-day, `Saturday, August 12, 2017 9:00 AM - Sunday, August 13, 2017 1:00 PM`. I first check for the multi-day case, and if that fails (i.e. when you can't parse an entire datetime from just a time), it combines the date of the `start_time` and the time of the `end_time`.

dispatch/apps/events/sources.py
```diff
@@ -1,22 +1,14 @@
...
-import datetime
-from HTMLParser import HTMLParser
+from datetime import datetime
 
 from django.conf import settings
 
 from dispatch.vendor.apis import Facebook, FacebookAPIError
 
-FACEBOOK_ERROR_MESSAGE = 'Invalid url: The event could be private, or there could be an error in the url itself. Check that the event is "public" and try again'
+from bs4 import BeautifulSoup
```
psiemens: Move this above the django imports

dispatch/apps/events/sources.py
```diff
@@ -41,15 +33,15 @@ def get_event_id(self, url):
         if m:
             return m.group(1)
         else:
-            raise FacebookEventError('URL provided is not a valid facebook event url')
+            raise EventError('URL provided is not a valid facebook event url')
```
psiemens: Capitalize "facebook"

dispatch/apps/events/views.py
```diff
@@ -60,3 +51,14 @@ def submit_form(request):
...
+def get_host_from_url(url):
+    """Parses URL against regex to pull out the hostname"""
+
+    # Match numbers that is the event id from the url and return them
+    m = re.search('.*(facebook\.com|calendar\.events\.ubc\.ca)+.*', url)
+
+    if m:
+        return m.group(1)
+    else:
+        raise EventError('URL provided is not a valid facebook event or UBC event url')
```
psiemens: Capitalize "facebook"

psiemens (changes requested):

psiemens (approved):

Issue #639, opened 2017-07-25
Article content is now too escaped


joncppl:
I was overzealous in https://github.com/ubyssey/dispatch/pull/627

now the built-in html editor doesn't work. (ie. hyperlinks, italic, bold, are broken)

perhaps the escape function can be modified to allow these tags while still escaping floating entities

PR #643, opened 2017-07-31
Selectively escape some html elements

joncppl:
Addresses #639

Since italic, bold, links, headers, etc used actual html, they need it to be unescaped.
Therefore blanket escaping all the html broke the editor.

It would be simple enough to revert that PR and restore the editor's functionality, but I think it is still worth while to try to escape "stray" html entities.

This PR escapes just two entities: `&` and `<`. But `<` is only escaped when it isn't followed by a valid html tag such as `a`, `em`, etc. This is done using a negative lookahead regex.

By escaping the `<`, a floating `>` becomes a non issue.

Now if a user inserts `<random text between angle brackets>`, it doesn't vanish, as the `<` is escaped, and the `>` alone doesn't indicate a tag, so it's a non issue.

Excuse my verbosity, I wanted to read it back myself to make sure it was sane.

### DIFF
--- a/dispatch/static/manager/package.json
+++ b/dispatch/static/manager/package.json
@@ -19,7 +19,6 @@
     "css-loader": "^0.26.1",
     "draft-js": "^0.10.0",
     "es6-promise": "^3.2.1",
-    "escape-html": "^1.0.3",
     "eslint": "^3.17.1",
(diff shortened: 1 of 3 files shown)


psiemens (approved):
Nice, this looks good to me 👍

PR #644, opened 2017-08-01
Redirect user to login screen if their token can not be verified

joncppl:
1. Add a token verification endpoint
2. check it when the app loads, if the user's token is invalid, redirect them to the login page. The token is not deleted from cookies, in case the request failed for other reasons, ie. internet failure. This way the user can just reload the page.

### DIFF
--- a/dispatch/apps/api/views.py
+++ b/dispatch/apps/api/views.py
@@ -458,8 +458,8 @@ class TokenViewSet(viewsets.ModelViewSet):
     def retrieve(self, request, pk):
 
         token_provided = pk
-        get_object_or_404(Token, key=token_provided, user=request.user)
-        return Response(status=status.HTTP_204_NO_CONTENT)
+        get_object_or_404(Token, key=token_provided)
+        return Response({'token-valid': True})
 
     def delete(self, request):
         token = get_object_or_404(Token, user=request.user)
--- a/dispatch/apps/api/views.py
+++ b/dispatch/apps/api/views.py
@@ -459,7 +459,7 @@ class TokenViewSet(viewsets.ModelViewSet):
         try:
             token = Token.objects.get(key=pk)
         except Token.DoesNotExist:
-            return Response({'token_valid': False}, status=status.HTTP_400_BAD_REQUEST)
+            return Response({'token_valid': False}, status=status.HTTP_404_NOT_FOUND)
 
         return Response({'token_valid': True})
 
--- a/dispatch/apps/api/views.py
+++ b/dispatch/apps/api/views.py
@@ -456,8 +456,11 @@ class TokenViewSet(viewsets.ModelViewSet):
             raise BadCredentials()
 
     def retrieve(self, request, pk):
+        try:
+            token = Token.objects.get(key=pk)
+        except Token.DoesNotExist:
+            return Response({'token_valid': False}, status=status.HTTP_400_BAD_REQUEST)
 
-        get_object_or_404(Token, key=pk)
         return Response({'token_valid': True})
 
     def delete(self, request):
@@ -466,3 +469,6 @@ class TokenViewSet(viewsets.ModelViewSet):
         token.delete()
 
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def get_queryset(self):
+        return None
--- a/dispatch/apps/api/views.py
+++ b/dispatch/apps/api/views.py
@@ -455,6 +455,12 @@ class TokenViewSet(viewsets.ModelViewSet):
         else:
             raise BadCredentials()
 
+    def retrieve(self, request, pk):
+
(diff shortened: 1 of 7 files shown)


joncppl:
oops, forgot to add a unit test for the new endpoint.

joncppl:
@psiemens ready for review now

dispatch/apps/api/views.py
```diff
@@ -455,6 +455,12 @@ def create(self, request):
         else:
             raise BadCredentials()
 
+    def retrieve(self, request, pk):
+
+        token_provided = pk
```
psiemens: I would just rename the `pk` argument to `token` and avoid the extra variable

dispatch/apps/api/views.py
```diff
@@ -455,6 +455,12 @@ def create(self, request):
...
+        get_object_or_404(Token, key=token_provided)
+        return Response({'token-valid': True})
```
psiemens: Change to `'valid_token'`

dispatch/static/manager/src/js/reducers/AuthReducer.js
```diff
@@ -38,4 +39,12 @@ reducer.handle(fulfilled(types.AUTH.DELETE_TOKEN), () => {
   return initialState
 })
 
+reducer.handle(fulfilled(types.AUTH.VERIFY_TOKEN), (state) => {
+  return R.merge(state, { validToken: true })
+})
+
+reducer.handle(rejected(types.AUTH.VERIFY_TOKEN), (state) => {
+  return R.merge(state, { validToken: false })
```
psiemens: Would it be possible to simply *remove* the token altogether, instead of using an extra `validToken` variable to keep track of its validity? I think if the server says the token is invalid, then the frontend should immediately remove it from the browser. Thoughts?

psiemens (changes requested):

dispatch/static/manager/src/js/reducers/AuthReducer.js
joncppl: does rejected include timeouts/other connection failures or only 4xx errors returned by the server? I don't like the idea of kicking people out of the app because their internet is inconsistent.

psiemens: I believe it includes timeouts and connection errors as well. That's a good point - definitely don't want to intermittently kick people off the app.

Could you verify using the boolean returned by the API?

joncppl: can do

joncppl: I can verify that the rejected handler runs when DNS lookup fails, so I can only assume it also does for timeouts. (for which I can't find a good way to simulate)

dispatch/apps/api/views.py
```diff
@@ -455,6 +455,11 @@ def create(self, request):
         else:
             raise BadCredentials()
 
+    def retrieve(self, request, pk):
+
+        get_object_or_404(Token, key=pk)
+        return Response({'token_valid': True})
```
psiemens: `token_valid` should be`False` if Token lookup fails

dispatch/static/manager/src/js/containers/AppContainer.js
```diff
@@ -6,13 +6,19 @@ import * as userActions from '../actions/UserActions'
...
-    if (!this.isAuthenticated()) {
-      this.props.onRequireLogin('/')
+    this.checkAuthenticated()
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.validToken != this.props.validToken) {
+      if (this.props.validToken === false) {
+        this.props.onRequireLogin('/')
+      }
```
psiemens: I think we can handle this logic directly in the `verifyToken` action with:

```js
dispatch(userActions.requireLogin(nextPath))
```

This would remove the need for the `validToken` store variable (I think)

dispatch/static/manager/src/js/actions/UserActions.js
```diff
@@ -35,6 +35,23 @@ export function authenticateUser(email, password, nextPath = '/') {
...
 
+export function verifyToken(token) {
+  return (dispatch) => {
+    dispatch({ type: pending(types.AUTH.VERIFY_TOKEN )})
+
+    return DispatchAPI.auth.verifyToken(token)
+      .then(() => {
+        dispatch({
+          type: fulfilled(types.AUTH.VERIFY_TOKEN)
+        })
```
psiemens: Include an `isValid` attribute in the action that can be used by the reducer to delete the cookie if necessary

psiemens (changes requested):

psiemens (approved):

PR #645, opened 2017-08-01
add more columns to the event index page

joncppl:
I think events got missed when I made the columns extendable since they were in a different branch at the time.

Anyway, pretty straightforward. 

Are there any particular columns that make more or less sense to display?


dispatch/static/manager/src/js/pages/Events/EventIndexPage.js
```diff
@@ -61,6 +62,13 @@ function EventsPageComponent(props) {
       typePlural='events'
       displayColumn='title'
       pageTitle='Events'
+      headers={[ 'Title', 'Published', 'Time', 'Location', 'Type']}
+      extraColumns={[
+        event => event.is_published ? 'Published' : 'Unpublished',
```
psiemens: Change 'Unpublished' to 'Not published'

psiemens (changes requested):

PR #648, opened 2017-08-02
Widget templatetag keyword arguments

joncppl:
The template tag for rendering zones accepts keyword arguments.

These are passed to the widgets render method. If the keywords are in the list of accepted keywords for that widget, that key/value is extended onto the context of the widget before it is rendered.

This data path adds flexibility to the theme, allowing data to handed between the view/widget context, reducing the need for redundant db queries.


psiemens (changes requested):
Great addition. 

The only concern I have is with the naming of the variables -- I'd prefer something a little less verbose.

PR #651, opened 2017-08-08
WidgetField (nested widgets)

joncppl:
Finish implementation of WidgetField, enabling nested widgets.

notes:
- Zones can have a `hidden` attribute, preventing them from showing  up in the main list of zones
- WidgetField takes a zone_id, specifying the compatible zone
- Had to reorganize reducer to allow multiple lists of widgets (sorted by zone) (as multiple WidgetSelectInputs can be displayed at once)
- Nested widget field info is 'injected' into the widget's json data.
- recursive normalization (normalizes nested events, articles, etc. but not widgets so as to maintain the hierarchy)
- It is also recursively strips out the widget field data before making a POST request so that it isn't stored in the database
- WidgetField validates all of it's child fields, and encodes the error messages accordingly.

### DIFF
--- a/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
+++ b/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
@@ -29,7 +29,7 @@ export default function WidgetField(props) {
         label={props.field.label}
         many={props.field.many}
         data={props.data}
-        zone_id={props.field.zone_id}
+        cWidgets={props.field.widgets}
         onChange={props.onChange} />
     </FormInput>
   )
--- a/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
+++ b/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
@@ -26,10 +26,8 @@ export default function WidgetField(props) {
       label={props.field.label}>
       <Field
         errors={childErrors}
-        label={props.field.label}
-        many={props.field.many}
+        field={props.field}
         data={props.data}
-        compatibleWidgets={props.field.widgets}
         onChange={props.onChange} />
     </FormInput>
   )
--- a/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
+++ b/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
@@ -7,25 +7,25 @@ import { FormInput } from '../inputs'
 export default function WidgetField(props) {
   const Field = fields[props.field.type]
 
-  let error = ''
-  let errors = null
+  let fieldError = ''
+  let childErrors = null
   if (props.field.type == 'widget' && props.error) {
     try {
-      errors = JSON.parse(props.error)
-      error = errors.self
+      childErrors = JSON.parse(props.error)
+      fieldError = childErrors.self
     } catch (e) {
-      error = props.error
+      fieldError = props.error
     }
   } else {
-    error = props.error
+    fieldError = props.error
   }
 
   return (
     <FormInput
-      error={error}
+      error={fieldError}
       label={props.field.label}>
       <Field
-        errors={errors}
+        errors={childErrors}
         label={props.field.label}
         many={props.field.many}
         data={props.data}
--- a/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
+++ b/dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
@@ -10,8 +10,12 @@ export default function WidgetField(props) {
   let error = ''
   let errors = null
   if (props.field.type == 'widget' && props.error) {
-    errors = JSON.parse(props.error)
-    error = errors.self
+    try {
+      errors = JSON.parse(props.error)
+      error = errors.self
+    } catch (e) {
+      error = props.error
+    }
   } else {
     error = props.error
   }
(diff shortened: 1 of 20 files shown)


dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
```diff
@@ -7,14 +7,29 @@ import { FormInput } from '../inputs'
 export default function WidgetField(props) {
   const Field = fields[props.field.type]
 
+  let error = ''
+  let errors = null
```
psiemens: `errors` defined twice

dispatch/static/manager/src/js/components/ZoneEditor/index.js
```diff
@@ -30,8 +30,33 @@ class ZoneEditorComponent extends React.Component {
     )
   }
 
+  processZone(zone) {
+    if (!zone) {
+      return zone
+    }
+
+    function processWidget(widget, fields) {
```
psiemens, line 38: Just to clarify, `processWidget` just removes the 'widget' property from each nested widget, right?

dispatch/static/manager/src/js/components/fields/WidgetField.js
```diff
@@ -0,0 +1,90 @@
...
+import WidgetFieldWrapper from '../ZoneEditor/WidgetField'
+
+class WidgetFieldComponent extends React.Component {
+  handleWidgetChange(widgetId) {
+    const data = R.merge(this.props.data || {}, {
+      id: widgetId,
+      data: this.props.data || {}
+    })
+
+    this.props.onChange(data)
```
psiemens: Could this method be re-written like this?

```js
this.props.onChange({
  id: widgetId,
  data: {}
})
```

In my opinion, changing the widget should clear the field data

dispatch/theme/fields.py
```diff
@@ -176,16 +200,55 @@ class EventField(ModelField):
...
     serializer = EventSerializer
 
-
 class WidgetField(Field):
 
     type = 'widget'
 
+    def __init__(self, label, zone, required=False):
+        super(WidgetField, self).__init__(label, required=required)
+        self.zone_id = zone.id
```
psiemens: How would you feel about passing in a list of widgets instead of associating widget fields with a "nested" zone? If all we are really trying to do is specify a list of compatible widgets for the field, I don't think we have to use a zone.

Would this add a lot of complexity to the current design?

psiemens (changes requested):

dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
joncppl: intentional. 
one is for the widget field itself and the other the array of children field's errors.
I'll rename to clarify.

dispatch/static/manager/src/js/components/ZoneEditor/index.js
joncppl: indeed.
should I make that clear via the function name or inline comment?

dispatch/static/manager/src/js/reducers/WidgetsReducer.js
```diff
@@ -15,7 +15,8 @@ const initialState = {
   isLoaded: false,
   selected: [],
   count: null,
-  ids: []
+  ids: [],
+  zones: {}
```
psiemens: Is this attribute still needed?

dispatch/static/manager/src/js/components/ZoneEditor/WidgetField.js
```diff
@@ -7,14 +7,29 @@ import { FormInput } from '../inputs'
...
     <FormInput
-      error={props.error}
+      error={fieldError}
       label={props.field.label}>
       <Field
+        errors={childErrors}
         label={props.field.label}
         many={props.field.many}
         data={props.data}
+        cWidgets={props.field.widgets}
```
psiemens: Maybe instead of passing `label`, `many` and `cWidgets`, we could just pass the entire `field` object?

dispatch/static/manager/src/js/components/inputs/WidgetSelectInput.js
```diff
@@ -18,12 +22,15 @@ class WidgetSelectInputComponent extends React.Component {
   }
 
   render() {
+    const results = this.props.cWidgets ? R.keys(this.props.cWidgets) : this.props.widgets.zones[this.props.zoneId]
```
psiemens: I'd prefer to be a bit more verbose and call it something like `this.props.compatibleWidgets`

dispatch/static/manager/src/js/components/fields/WidgetField.js
```diff
@@ -0,0 +1,71 @@
...
+    const widgetData = {}
+    widgetData[name] = data
+
+    this.props.onChange(R.merge(this.props.data || {}, {
+      data: R.merge(this.props.data.data || {}, widgetData)
+    }))
+  }
+
+  getWidgetId() {
+    return R.prop('id', this.props.data || {})
```
psiemens: You can make use of [default props](https://facebook.github.io/react/docs/typechecking-with-proptypes.html#default-prop-values) to get rid of all the `this.props.data || {}` expressions throughout this class - makes things a bit more readable.

psiemens (changes requested):
This is really close -- just a few remaining comments!

PR #655, opened 2017-08-09
Widget embed type

psiemens:


dispatch/apps/frontend/embeds.py
```diff
@@ -186,6 +186,16 @@ def prepare_data(self, data):
             'size': len(images)
         }
 
+class WidgetController(AbstractController):
+
+    @classmethod
+    def render(self, data):
+
```
psiemens, line 194: The render method should return the result of rendering the widget to HTML.

```py
widget = ThemeManager.Widgets.get(data['widget_id'])
widget.set_data(data['data'])

return widget.render()

File: dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/WidgetEmbed.js [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+          <TextInput
+            fill={true}
+            value={props.data.content}
+            onChange={e => props.updateField('content', e.target.value)} />
+
+        </FormInput>
+
+        <FormInput>
+          <WidgetSelectInput
+            zoneId={props.data.zoneId}
```
psiemens: - `zoneId` is always 'embed'
- the current widget ID should be stored in `data.widget_id`
```jsx
<WidgetSelectInput
  zoneId='embed' 
  selected={props.data.widget_id}
  update={widgetId => props.updateField('widget_id', widgetId)} />

File: dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/WidgetEmbed.js [hunk 3]
```diff
@@ -0,0 +1,41 @@
...
+        </FormInput>
+
+    </div>
+  )
+}
+
+export default {
+  type: 'widget',
+  component: WidgetEmbedComponent,
+  defaultData: {
```
psiemens, line 40: `defaultData` should be structured like this:

```js
defaultData: {
  widgetId: null,
  data: {}
}

PR #656, opened 2017-08-10
ImageInput should load selected images when mounted

psiemens:
I noticed that `ImageInput` does not fetch selected images when first mounted on the page. This PR fixes that

### DIFF
--- a/dispatch/static/manager/src/js/components/inputs/ImageInput.js
+++ b/dispatch/static/manager/src/js/components/inputs/ImageInput.js
@@ -5,6 +5,7 @@ import { connect } from 'react-redux'
 import { AnchorButton } from '@blueprintjs/core'
 
 import * as modalActions from '../../actions/ModalActions'
+import imagesActions from '../../actions/ImagesActions'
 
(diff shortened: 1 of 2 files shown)


joncppl (approved):

Axel-Jacobsen (approved):

dispatch/static/manager/src/js/components/inputs/ImageInput.js
```diff
@@ -24,6 +25,10 @@ function Image(props) {
 
 class ImageInputComponent extends React.Component {
 
+  componentDidMount() {
+    this.getSelected().map(id => this.props.getImage(this.props.token, id))
```
psiemens: I just realized that this could be a bit cleaner if it was moved to its own method:

```js
loadSelectedImages() {
  this.getSelected().map(id => this.props.getImage(this.props.token, id))
}
```

@Axel-Jacobsen @joncppl can one of you make this change for me?

psiemens:
Thanks @joncppl -- merging now

PR #657, opened 2017-08-10
Changed url for event api from `event` to `events`

Axel-Jacobsen:
Takes care of #653


joncppl:
I think the react app needs to reflect the change in the api url.

https://github.com/ubyssey/dispatch/blob/develop/dispatch/static/manager/src/js/api/dispatch.js#L352

psiemens (approved):

PR #658, opened 2017-08-10
Added validator to make sure that we have start/end times

Axel-Jacobsen:
We had a validator to confirm that the start time was before the end time, but neither the start nor end times were confirmed to exist before this. Added validators and made start/end times required in Event Model


psiemens (approved):
Looks good to me!

PR #665, opened 2017-08-16
Profile page

joncppl:
For the users api route, I added a feature insofar that an authenticated user can provide a keyword instead of a user id to get the authenticated user data. Since there is currently no way for the frontend to know the authenticated user's id. I chose the keyword `me` but it could be anything.

Added actions & reducer for the user model.

For the profile page itself, I reused the ProfileForm component.
Changing user name and password is an independent form with a separate submit button (since patching user is a different api call). I think this is fairly common on websites. 

There is currently no way to change the person referenced by the user, only update the person already associated. Do we want to make it changeable?

### DIFF
--- a/dispatch/static/manager/src/js/pages/ProfilePage.js
+++ b/dispatch/static/manager/src/js/pages/ProfilePage.js
@@ -7,6 +7,8 @@ import userActions from '../actions/userActions'
 import PersonEditor from '../components/PersonEditor'
 import { FormInput, TextInput } from '../components/inputs'
 
+require('../../styles/components/user_form.scss')
+
 class ProfilePageComponent extends React.Component {
   loadUser() {
     this.props.getUser(this.props.token, 'me')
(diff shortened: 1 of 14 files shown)


dispatch/static/manager/src/js/pages/ProfilePage.js
```diff
@@ -1,7 +1,141 @@
...
+        <form onSubmit={e => e.preventDefault()}>
+          <FormInput
+            label='Email'
+            padded={false}
+            error={errors.email}>
+            <TextInput
+              placeholder='name@domain.tld'
+              value={user.email || ''}
+              fill={true}
+              onChange={ e => this.handleUpdate('email', e.target.value) } />
```
psiemens: Remove space before `e`

psiemens (approved):

psiemens (changes requested):
Before merging, can you add a simple test case for the new `retrieve` method?

psiemens (approved):

PR #675, opened 2017-08-21
Style the dispatch login form

joncppl:
Also pass errors from server through to be displayed. (ie. invalid credentials)


dispatch/static/manager/src/js/pages/LoginPage.js
```diff
@@ -32,19 +34,35 @@ class LoginPageComponent extends React.Component {
...
+          </div>
           <input
+            className='pt-input pt-round'
             name="email"
             type="email"
+            placeholder='Email'
             value={this.state.email}
             onChange={this.onChangeEmail} /><br/>
           <input
+            className='pt-input pt-round'
```
psiemens: I'm not a huge fan of the rounded corners -- just seems a bit *too* round to me

dispatch/static/manager/src/js/pages/LoginPage.js
```diff
@@ -32,19 +34,35 @@ class LoginPageComponent extends React.Component {
...
             name="password"
             type="password"
+            placeholder='Password'
             value={this.state.password}
             onChange={this.onChangePassword} /><br/>
-          <input type="submit" />
+          <button
+            className='pt-button pt-large'
```
psiemens: The button seems a bit big compared to the text inputs -- maybe just use the regular size?

dispatch/static/manager/src/js/pages/LoginPage.js
```diff
@@ -32,19 +34,35 @@ class LoginPageComponent extends React.Component {
 
   render() {
     return (
-      <div>
-        <form onSubmit={this.onSubmit}>
+      <div className='c-login-page'>
+        <form onSubmit={this.onSubmit} className='pt-card pt-elevation-2'>
+          <div className='c-login-page__heading'>
+            Dispatch Login
```
psiemens: I think we can just say "Dispatch"

psiemens (changes requested):
Looks good and works really well -- the error messages are great.

Are you able to make the login box a bit wider? The text fields just seem a bit squished, especially for longer email addresses.

psiemens (approved):

PR #689, opened 2017-08-29
Documentation

joncppl:
Here is the the documentation I wrote last week. 
I'm afraid, It's not complete but it does cover many points.

Hope it helps :)


psiemens:
Thanks!!

PR #706, opened 2017-09-27
Extract CSS into separate stylesheet bundle

psiemens:
This modifies the current webpack configuration to create a separate CSS bundle, allowing us to better leverage browser stylesheet caching.

@JamieRL review plzzz


JamieRL (approved):
Looks good to me. One question, why the name change to manager-[version]?

psiemens:
@JamieRL I just thought it made more sense, especially since `index.js` often ended up in the same bucket folder as other static files so it wasn't obvious what its purpose was.

PR #716, opened 2017-10-05
Make embed saving/loading more robust

psiemens:


dispatch/api/loaders.py
```diff
@@ -0,0 +1,42 @@
+from dispatch.modules.content.models import Image, ImageGallery
+
+class Loader(object):
+    """A Loader is used to efficiently fetch data for content embeds.
+
+    Each loader has a corresponding embed type, and facilitates the prefecting
```
blx: `prefecting` -> `prefetching` ?

dispatch/api/serializers.py
```diff
@@ -279,26 +283,62 @@ def to_internal_value(self, content):
...
-        result = []
-        for block in content:
-            if block['type'] == 'image':
-                block['data']['image'] = ImageSerializer(self.images[block['data']['image_id']]).data
-            if block['type'] == 'gallery':
-                block['data']['gallery'] = ImageGallerySerializer(self.galleries[block['data']['id']]).data
-            result.append(block)
-        return result
+        """Insert loaded data into embed data blocks."""
+        return map(lambda block: self.insert_instance(block), content)
```
blx: could be
```py
return map(self.insert_instance, content)
```

dispatch/modules/content/render.py
```diff
@@ -11,7 +11,10 @@ def render_node(html, node):
         if node['type'] == 'paragraph':
             return html + '<p>%s</p>' % node['data']
         else:
-            return html + embedlib.render(node['type'], node['data'])
+            try:
+                return html + embeds.render(node['type'], node['data'])
+            except EmbedException:
+                return html
```
blx, line 17: can this be logged somewhere (eg. as a warning into a log file or sentry)?

psiemens: Yeah, was thinking that -- meant to look into logging methods.

I think we can just use Python's built-in `logging` module and let the user configure their own handlers.

PR #718, opened 2017-10-07
Fix SelectInput so that placeholder is default selection (fixes #717)

### LINKED ISSUE #717: 'Category' bug on the New Event page in the Admin Panel.
When the form loads, the default category selected/shown is "Sports".

However, if you leave the category as "Sports" and try to save the form, you will get an error message saying `"" is not a valid choice.`

In order to actually select "Sports" as your event's category, you have to select some other random category from the dropdown box first, and then select "Sports" again.

Meanwhile, I see that there is a "Please Select" option in the dropdown box that was probably intended to be the placeholder/default selection instead of "Sports"?

i-ron-y:
Close #717


psiemens (approved):
Nice catch! Thanks for the fix

PR #724, opened 2017-10-24
Created runtime slug validator

bf777:
Contains slug validator, updated slug serializer, and new slug test.

### DIFF
--- a/dispatch/api/serializers.py
+++ b/dispatch/api/serializers.py
@@ -8,7 +8,7 @@ from dispatch.modules.content.models import (
 from dispatch.modules.auth.models import Person, User
 
 from dispatch.api.mixins import DispatchModelSerializer, DispatchPublishableSerializer
-from dispatch.api.validators import ValidFilename, ValidateImageGallery, PasswordValidator
+from dispatch.api.validators import ValidFilename, ValidateImageGallery, PasswordValidator, SlugValidator
 from dispatch.api.fields import JSONField, PrimaryKeyField, ForeignKeyField
 
 from dispatch.theme.exceptions import WidgetNotFound, InvalidField
@@ -306,6 +306,7 @@ class ArticleSerializer(DispatchModelSerializer, DispatchPublishableSerializer):
     """
 
     id = serializers.ReadOnlyField(source='parent_id')
+    slug = serializers.SlugField(validators=[SlugValidator()])
 
     section = SectionSerializer(read_only=True)
     section_id = serializers.IntegerField(write_only=True)
--- a/dispatch/api/serializers.py
+++ b/dispatch/api/serializers.py
@@ -424,6 +424,7 @@ class PageSerializer(DispatchModelSerializer, DispatchPublishableSerializer):
     """
 
     id = serializers.ReadOnlyField(source='parent_id')
+    slug = serializers.SlugField(validators=[SlugValidator()])
 
     featured_image = ImageAttachmentSerializer(required=False, allow_null=True)
 
--- a/dispatch/api/validators.py
+++ b/dispatch/api/validators.py
@@ -37,7 +37,6 @@ def ValidateImageGallery(data):
             raise InvalidGalleryAttachments('One or more image ids were not provided')
 
 class SlugValidator(object):
-
     def set_context(self, serializer_field):
         self.instance = serializer_field.parent.instance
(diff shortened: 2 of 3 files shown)


psiemens:
@bf777 your test is just missing the case when you are trying to change the slug of an _already saved_ article to an existing slug of another article.

https://codecov.io/gh/ubyssey/dispatch/pull/724/src/dispatch/api/validators.py

dispatch/tests/test_api_articles.py
```diff
@@ -61,6 +61,74 @@ def test_create_article(self):
         self.assertEqual(response.data['authors'][0]['full_name'], 'Test Person')
         self.assertEqual(response.data['slug'], 'test-article')
 
+    def test_check_slug(self):
+        """Ensure that article doesn't have slug matching existing article"""
+
+        response = DispatchTestHelpers.create_article(self.client)
```
psiemens: I would create the articles like this:

```py
article_a = DispatchTestHelpers.create_article(self.client, slug='slug-a')
article_b = DispatchTestHelpers.create_article(self.cleint, slug='slug-a')

# article_b should return 400 error
self.assertEqual(article_b.status_code, status.HTTP_400_BAD_REQUEST)

File: dispatch/api/validators.py [hunk 2]
```diff
@@ -35,3 +35,17 @@ def ValidateImageGallery(data):
             raise InvalidGalleryAttachments('One or more images does not exist')
         except KeyError:
             raise InvalidGalleryAttachments('One or more image ids were not provided')
+
+class SlugValidator(object):
+
```
psiemens: Remove this empty line

dispatch/api/serializers.py
```diff
@@ -306,6 +306,7 @@ class ArticleSerializer(DispatchModelSerializer, DispatchPublishableSerializer):
     """
 
     id = serializers.ReadOnlyField(source='parent_id')
+    slug = serializers.SlugField(validators=[SlugValidator()])
```
psiemens, line 309: Can you also add `SlugValidator` to `PageSerializer`?

psiemens (changes requested):

dispatch/tests/test_api_articles.py
```diff
@@ -64,70 +64,17 @@ def test_create_article(self):
...
-        }
+        #article_b should return 400 error
+        self.assertEqual(article_b.status_code, status.HTTP_400_BAD_REQUEST)
 
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-
-        # Now check that slug can't be changed to slug of existing article
-
-        article = DispatchTestHelpers.create_article(self.client)
+        article_c = DispatchTestHelpers.create_article(self.client, slug='slug-b')
```
psiemens: Try this and see if it works for the second case:

```py
article_a = DispatchTestHelpers.create_article(self.client, slug='slug-a')
article_b = DispatchTestHelpers.create_article(self.client, slug='slug-b')

url = reverse('api-articles-detail', args=[article_b.data['id']])
data = {'slug': 'slug-a'}

# Update `article_b` with `slug-a`
response = self.client.patch(url, data, format='json')

self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)

**Review by psiemens — CHANGES_REQUESTED** [review]

File: dispatch/tests/test_api_articles.py [hunk 5]
```diff
@@ -61,6 +61,20 @@ def test_create_article(self):
         self.assertEqual(response.data['authors'][0]['full_name'], 'Test Person')
         self.assertEqual(response.data['slug'], 'test-article')
 
+    def test_check_slug(self):
+        """Ensure that article doesn't have slug matching existing article"""
+
```
psiemens: Ah my bad -- I should have been more clear earlier!

The code below is good, but instead of replacing the old test, we actually need to create two cases:
1. Creating a **new** article with the slug of on existing article should fail
2. Updating an **existing** article with the slug of an existing article should fail

It might make sense to write these as two separate test methods:

```py
# Case 1
def test_create_article_existing_slug(self):
  ...

# Case 2
def test_update_article_existing_slug(self):
  ...
```

Case 2 is already implemented below, so all you'll have to do is add the first one back.

psiemens (changes requested):

PR #725, opened 2017-10-26
Added facebook and twitter fields to the person form

BenGHolmes:


BenGHolmes:
Just realized I accidentally hard coded something on the front end that will need to be implemented in dispatch. I'll update and commit again tomorrow

dispatch/migrations/0001_initial.py
```diff
@@ -0,0 +1,293 @@
+# -*- coding: utf-8 -*-
```
psiemens: Since we haven't been committing migrations files (although we should maybe start), can you remove this file?

dispatch/static/manager/src/js/api/dispatch.js
```diff
@@ -259,6 +259,7 @@ const DispatchAPI = {
       return getRequest('persons', personId, null, token)
     },
     save: (token, personId, data) => {
+      console.log(data.values())
```
psiemens: Remove `console.log`

psiemens (changes requested):
Looks solid! Just a few small changes before it's good to go.

psiemens (approved):

psiemens:
Feel free to merge once the tests pass 🤘

PR #732, opened 2017-10-31
697 articles filter by section

JamieRL:
Adding functionality to filter articles by author and section. Re-used AuthorSelectInput and SectionSelectInput with slight modification as to not show the search box default props keep same appearance in the article editor, modified searchArticles action to add authors. Can only filter by 1 author at a time.


JamieRL:
Looks like there's an issue with migrations causing the travis build to fail

psiemens:
@JamieRL try updating from `develop` -- I patched the migration files

JamieRL:
Looks like that fixed it. 

I'll add a test for searching by author.

JamieRL:
Test for api search articles by author added.

JamieRL:
I noticed that if you were to navigate manually to /articles/?section=1 for example that it there's an error rendering the filter button.

dispatch/tests/test_api_articles.py
```diff
@@ -476,3 +476,23 @@ def test_section_name_query(self):
...
         self.assertEqual(data['results'][1]['headline'], 'Article 1')
         self.assertEqual(data['count'], 2)
+
+    def test_author_name_query(self):
+        """Should be able to search for articles by author name"""
+
+        article_1 = DispatchTestHelpers.create_article(self.client, headline='Article 1', slug='article-1')
+        article_2 = DispatchTestHelpers.create_article(self.client, headline='Article 2', slug='article-2')
+
+        person = Person.objects.create(full_name='Test Person2')
```
psiemens: This line is a bit redundant -- I think you can just pass the `full_name` to `DispatchTestHelpers.create_article` and it will handle the person creation.

```py
article_3 = DispatchTestHelpers.create_article(self.client, headline='Article 3', slug='article-3', author='Test Person2')
```

dispatch/api/views.py
```diff
@@ -71,13 +71,17 @@ def get_queryset(self):
...
+        author = self.request.query_params.get('author', None)
 
         if q is not None:
             queryset = queryset.filter(headline__icontains=q)
 
         if section is not None:
             queryset = queryset.filter(section_id=section)
 
+        if author is not None:
+            queryset = queryset.filter(authors__id=author)
```
psiemens, line 83: Have you tested this when an article has multiple authors? Might be worth adding a test case.

psiemens (changes requested):
The backend changes look solid. Just wondering about the behaviour when an article has multiple authors.

The frontend code looks and works great, but I'm thinking there might be a slightly cleaner way to separate the filter logic from the normal select inputs -- possibly with a separate component? I'm not sure if that would add unneeded complexity though. I'll follow up with another review ASAP

dispatch/api/views.py
JamieRL: Didn't test it in the test case but when testing it out on the front end filtering by author=authorA catches articles with authors=[authorA, authorB]. I'll add a case.

JamieRL:
I took a shot at making a separate filter component but I'm not sure it's any better 

@psiemens let me know what you think/if this is what you meant

JamieRL:
@psiemens looks like adding a remove button to the button would be a little complicated. Is this good to go as is?

dispatch/static/manager/src/js/components/inputs/AuthorSelectInput.js
```diff
@@ -17,16 +17,39 @@ class AuthorSelectInputComponent extends React.Component {
...
+    if (isFilter) {
+      return this.getCurrentFilter(selected)
+    }
+
+    if (selected) {
+      return 'Edit Authors'
+    }
+
+    return 'Add Authors'
+  }
```
psiemens: Are these additions still needed?

dispatch/static/manager/src/js/components/inputs/SectionSelectInput.js
```diff
@@ -17,17 +17,40 @@ class SectionSelectInputComponent extends React.Component {
     this.props.listSections(this.props.token, queryObj)
   }
 
+  getCurrentFilter(selected) {
+    if (this.props.entities.sections[selected]) {
+      return this.props.entities.sections[selected].name
+    }
+
+    return 'Filter by Section'
+  }
```
psiemens: Same here -- is this still needed?

psiemens (changes requested):
I think this is super close to being ready -- the only thing I'm hesitant about is this code that switches between filter types:

```js
fetchResults(query) {
  if (this.props.filterBy === 'sections') {
    this.listItems(query, this.props.listSections)
  }
  else if (this.props.filterBy === 'authors') {
    this.listItems(query, this.props.listAuthors)
  }
}
```

It works for now but isn't very flexible if we want to add more filters in the future. I'm thinking we could instead add `AuthorFilterInput` and `SectionFilterInput` components that act as simple wrappers around `FilterSelectInput`, similar to how `AuthorSelectInput` works: https://github.com/ubyssey/dispatch/blob/develop/dispatch/static/manager/src/js/components/inputs/AuthorSelectInput.js

@JamieRL What do you think?

dispatch/static/manager/src/js/components/inputs/SectionSelectInput.js
JamieRL: Right those shouldn't be needed anymore. I'll remove them.

dispatch/static/manager/src/js/components/inputs/filters/AuthorFilterInput.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    this.props.listPersons(this.props.token, queryObj)
+  }
+
+  render() {
+    return (
+      <FilterSelectInput
+        selected={this.props.selected}
+        filterBy='authors'
```
psiemens: Not needed anymore

dispatch/static/manager/src/js/components/inputs/filters/AuthorFilterInput.js
```diff
@@ -0,0 +1,60 @@
...
+        results={this.props.persons}
+        entities={this.props.entities.persons}
+        update={(selected) => this.props.update(selected)}
+        fetchResults={(query) => this.listPersons(query)}
+        attribute='full_name'
+        editMessage={this.props.selected ? this.props.entities['persons'][this.props.selected].full_name : 'Filter by Author'}
```
psiemens: can change `this.props.entities['persons']` to `this.props.entities.persons`

psiemens: also maybe move this logic into `FilterSelectInput`

dispatch/static/manager/src/js/components/inputs/filters/FilterSelectInput.js
```diff
@@ -0,0 +1,31 @@
...
+        editMessage={this.props.editMessage}
+        filterButton={true} />
+    )
+  }
+
+}
+
+
+
+
```
psiemens: Some extra space here ^^

dispatch/static/manager/src/js/components/inputs/filters/SectionFilterInput.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { connect } from 'react-redux'
+
+import FilterSelectInput from './FilterSelectInput'
+
+import sectionsActions from '../../../actions/SectionsActions'
+
+class SectionSelectInputComponent extends React.Component {
```
psiemens: should be `SectionFilterInputComponent`

dispatch/static/manager/src/js/pages/Articles/ArticleIndexPage.js
```diff
@@ -46,28 +49,45 @@ function ArticlePageComponent(props) {
   const section = props.entities.sections[props.location.query.section]
   const title = section ? `${section.name} - Articles` : 'Articles'
 
+  const filters = (
+    <div className='c-item-list__header__filters'>
```
psiemens: To simplify this a little bit, filters could be an array of components: 

```jsx
const filters = [
  <SectionFilterInput
    selected={props.location.query.section}
    update={(section) => props.searchArticles(props.location.query.author, section, props.location.query.q)}
    />,
  <AuthorFilterInput
    selected={props.location.query.author}
    update={(author) => props.searchArticles(author, props.location.query.section, props.location.query.q)}
    />
]
```

And then the `ItemListHeader` component could wrap them in the div:

```jsx
<div className='c-item-list__header__filters'>{filters}</div>
```

psiemens (changes requested):

psiemens (approved):

psiemens:
Looks good! Merging now 🤘

PR #739, opened 2017-11-07
Add preview functionality to article editor.

### LINKED ISSUE #708: Add preview functionality to article editor
Requirements:
- [x] Add a new `preview_id` field to the `Publishable` model
  - should be type `UUIDField`
  - randomly generated value for each version of a `Publishable` instance (`Article` or `Page`)
- [x] The article URL should accept two URL querystring parameters: `version` and `preview_id`
  - e.g. `/?version=4&preview_id=1a5h8a21812349`
  - If the `version` and `preview_id` match a row in the database for the given article, display that version of the article whether or not it has been published
  - to implement this you'll want to update this method: https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/helpers.py#L15-L17
- [x] The preview button in the article editor should open the preview url for the current version of the article/page in a new tab

i-ron-y:
Close #708


dispatch/modules/content/models.py
```diff
@@ -3,13 +3,17 @@
 import os
 import re
 
+import uuid
```
psiemens: This can be placed directly below `import re`

dispatch/modules/content/models.py
```diff
@@ -3,13 +3,17 @@
...
 from jsonfield import JSONField
 from PIL import Image as Img
 
 from django.db import IntegrityError
 from django.db.models import (
     Model, DateTimeField, CharField, TextField, PositiveIntegerField,
-    ImageField, FileField, BooleanField, ForeignKey, ManyToManyField,
+    ImageField, FileField, BooleanField,
+    UUIDField,
+    ForeignKey, ManyToManyField,
```
psiemens: I would put these on the same line as `UUIDField`

dispatch/static/manager/src/js/actions/PublishableActions.js
```diff
@@ -66,8 +66,8 @@ export default class PublishableActions extends ResourceActions {
               )
             }
           })
-
-          openPreviewWindow(json.url, json.id)
+          openPreviewWindow(json.url + '?version=' + json.current_version +
```
psiemens: You can use ES6 string templates to make this a bit cleaner:

```js
openPreviewWindow(
  `${json.url}?version={json.current_version}&preview_id=${json.preview_id}`, json.id
)

**Review by psiemens — CHANGES_REQUESTED** [review]
Overall functionality looks solid!

File: dispatch/api/serializers.py [hunk 4]
```diff
@@ -443,6 +444,7 @@ class PageSerializer(DispatchModelSerializer, DispatchPublishableSerializer):
     class Meta:
         model = Page
         fields = (
+            'preview_id',
```
psiemens: One more thing -- can you place this field after `latest_version`?

psiemens (changes requested):

psiemens (approved):

psiemens:
@i-ron-y Ah, I forgot one thing -- we should add `preview_id` to the `authenticated_fields` tuple so that unauthenticated users can never view the preview_id. Leaking the preview_id for an already published article isn't really a big deal, but it's better not to expose "secret" fields.

Here's where that should be added:
https://github.com/ubyssey/dispatch/blob/eba6ffea99a41a81416b1975a91eaae35b223e8e/dispatch/api/serializers.py#L374-L377

PR #741, opened 2017-11-07
Fixed Firefox image manager - broken flexbox

bf777:
Firefox requires a `min-height` and `min-width` setting for flexbox scrolling to work properly (apparently [Chrome might soon require this](https://stackoverflow.com/questions/28636832/firefox-overflow-y-not-working-with-nested-flexbox)  so it's good future-proofing).


psiemens:
Tested in Firefox 56.0 -- works great! Thanks @bf777

psiemens:
Tested in Firefox 56.0 -- works great! Thanks @bf777

psiemens (approved):

PR #742, opened 2017-11-14
Pending Events badge

BenGHolmes:


dispatch/static/manager/src/js/containers/MainContainer.js
```diff
@@ -28,7 +33,7 @@ class Main extends React.Component {
     return (
       <div>
         <Toaster className='c-toaster' position={Position.TOP} ref='toaster' />
-        <Header />
+        <Header pendingCount={this.props.pending}/>
```
psiemens: Add a space before the closing tag:
```jsx
<Header pendingCount={this.props.pending} />

File: dispatch/static/manager/src/styles/components/header.scss [hunk 2]
```diff
@@ -25,3 +25,30 @@
   box-shadow: none;
   border: none;
 }
+
+.event-badge{
```
psiemens: Can you change this classname to `.c-banner__badge`, just to make it a bit more reusable?

psiemens (changes requested):
This looks great! Nice job with the CSS styling.

Just a few small things to change before we can merge

psiemens (changes requested):
Looks solid!

However, the `header.min.css` file shouldn't be there. I think maybe it got added to the src folder accidentally? Just remove that and you're good to merge 🤘

PR #743, opened 2017-11-15
Move the logic from ArticleHelper to PublishableManager

i-ron-y:
Close ubyssey/ubyssey.ca#390

So I'm not entirely sure if this is what you had in mind, but I relocated all the logic (for deciding whether to get the published or the preview version of the article) from ArticleHelper in helpers.py (ubyssey.ca repo) to PublishableManager in managers.py (dispatch repo).

Let me know what you think!

### DIFF
--- a/dispatch/modules/content/managers.py
+++ b/dispatch/modules/content/managers.py
@@ -15,16 +15,17 @@ class PublishableManager(Manager):
 
         Otherwise, get the published version of the article.
         """
-        
-        request = args[0]
+        kwargs['is_published'] = True
 
-        version = request.GET.get('version', None)
-        preview_id = request.GET.get('preview_id', None)
+        if len(args) > 0:
+        	request = args[0]
 
-        if (version is not None) and (preview_id is not None):
-        	kwargs['revision_id'] = version
-        	kwargs['preview_id'] = preview_id    
-        else:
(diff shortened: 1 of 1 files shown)


i-ron-y:
Whoops, the build is failing... gotta try to figure this one out...

dispatch/modules/content/managers.py
```diff
@@ -9,4 +9,22 @@ def get(self, *args, **kwargs):
...
             del kwargs['pk']
-        return super(PublishableManager, self).get(*args, **kwargs)
+
+        """If the url requested includes the querystring parameters 'version' and 'preview_id',
+        get the article with the specified version and preview_id.
+
+        Otherwise, get the published version of the article.
+        """
+
+        if len(args) > 0:
```
psiemens: I think the build is failing because `request` is now a required parameter for `Article.objects.get`, so all existing calls to this method will fail.

To fix this, you can make `request` an optional keyword argument and call the method like this:

```python
article = Article.objects.get(request=request, slug=slug, is_published=True)
```

This _should_ avoid breaking changes to existing calls to `Article.objects.get`.

psiemens (changes requested):

psiemens (approved):

PR #744, opened 2017-11-27
Added a way for authors to be assigned new roles (photographer etc.)

### LINKED ISSUE #750: Add support for author types

bf777:
This new functionality will allow an author's type to be selected during the article creation process and used throughout a news website. This will allow photographers and illustrators to be more prominently featured in articles, and will later give them the opportunity to have their content (galleries etc.) featured in one centralized location on the website.

Close #750


dispatch/api/serializers.py
```diff
@@ -498,7 +513,7 @@ def update(self, instance, validated_data):
         if featured_image != False:
             instance.save_featured_image(featured_image)
 
-        authors = validated_data.get('author_ids')
+        authors = validated_data.get('authors')
```
psiemens: This should actually stay as "author_ids"

dispatch/modules/content/models.py
```diff
@@ -302,14 +302,23 @@ def save_topic(self, topic_id):
...
         Author.objects.filter(article=self).delete()
 
         # Create a new author for each person in list
         # Use `n` to save authors in correct order
-        for n, person_id in enumerate(authors):
-            Author.objects.create(article=self, person_id=person_id, order=n)
+        for n, author in enumerate(authors):
+            Author.objects.create(
+                article = self,
+                person_id = author['person["id"]'],
```
psiemens: This is close! If you remove the `person` argument and use `person_id=author['person']` it should work. It's a bit counterintuitive, I know.

`person_id` expects an integer ID and `person` expects an instance of a person. In this case, `author['person']` is an ID and not an instance, so we have to use the `person_id` argument. Both accomplish the same thing, though.

dispatch/api/validators.py
```diff
@@ -48,3 +48,11 @@ def __call__(self, slug):
         else:
             if self.model.objects.filter(slug=slug).exclude(parent=self.instance.parent).exists():
                 raise ValidationError('%s with slug \'%s\' already exists.' % (self.model.__name__, slug))
+
+class TypeValidator(object):
```
psiemens: This actually doesn't need to be a class. I would rename it to `AuthorValidator` and structure it like so:

```py
def AuthorValidator(data):
  # Check that data['person'] and data['type'] is set
```

dispatch/api/serializers.py
```diff
@@ -410,8 +425,8 @@ class ArticleSerializer(DispatchModelSerializer, DispatchPublishableSerializer):
 
     content = ContentSerializer()
 
-    authors = PersonSerializer(many=True, read_only=True)
-    author_ids = serializers.ListField(write_only=True, child=serializers.IntegerField())
+    authors = AuthorSerializer(source='get_authors', many=True, read_only=True)
+    author_ids = serializers.ListField(write_only=True, child=serializers.JSONField())
```
psiemens: Once `AuthorValidator` is defined, you can include it here in the `validators` argument

psiemens (changes requested):

dispatch/tests/helpers.py
```diff
@@ -15,9 +15,13 @@ def create_article(cls, client, headline='Test headline', slug='test-article', s
...
         # Create test person
         authors = []
-        for author in author_names:
-            (person, created) = Person.objects.get_or_create(full_name=author)
+
+        for person in author_names:
+            (person, created) = Author.objects.get_or_create(person=(
+            (Person.objects.get_or_create(full_name=person))
+            ),type='author')
             authors.append(person.id)
```
psiemens: Instead of appending the person ID, you'll want to append a dictionary that contains the person id **and** the author type:

```python
(person, created) = Person.objects.get_or_create(full_name=author)
authors.append({
  'person': person.id,
  'type': 'author'
})
```

This is the new format that the API expects after we updated the serializer.

dispatch/modules/content/models.py
```diff
@@ -356,6 +371,7 @@ class Author(Model):
     image = ForeignKey('Image', null=True)
     person = ForeignKey(Person)
     order = PositiveIntegerField()
+    type = CharField(blank=True, default='author', max_length=100)
```
psiemens, line 374: Try adding `null=True` to this as well

dispatch/api/validators.py
```diff
@@ -48,3 +48,15 @@ def __call__(self, slug):
...
+def AuthorValidator(data):
+    """Raise a ValidationError if data does not match the author format."""
+    if not isinstance(data, list):
+        # Convert single instance to a list
+        data = [data]
+
+    for author in data:
+        if 'person' not in author:
+            raise ValidationError('An author must contain a person.')
+        if 'type' in author and not isinstance('type', str):
```
psiemens: This should be:

```py
if 'type' in author and not isinstance(author['type'], str):
```

psiemens (approved):

PR #748, opened 2017-12-09
Video model

i-ron-y:


psiemens (changes requested):
This looks perfect 👌 

I know it's annoying, but can you add a simple test suite before merging? (similar to https://github.com/ubyssey/dispatch/blob/develop/dispatch/tests/test_api_tags.py)

psiemens (approved):
Good to go!

PR #752, opened 2017-12-26
Fix event approve button

psiemens:
A small fix for the event approval button -- it was being linked to the wrong action.


bf777 (approved):
Looks good! Thanks for fixing this up!

PR #754, opened 2018-01-10
700 zone search

JamieRL:
Added search by zone functionality to widget page. Hooked up same as article page etc. Search method is in ThemeManager, passes back to view which handles serialization of objects.


JamieRL:
Adding tests.

dispatch/static/manager/src/js/pages/Widgets/ZoneIndexPage.js
```diff
@@ -7,10 +7,37 @@ import ItemList from '../../components/ItemList'
...
+      query.q = this.props.location.query.q
+    }
+    return query
+  }
+
+  isNewQuery(prevProps, props) {
+    return prevProps.location.query.q !== props.location.query.q
+  }
+
+  handleSearchZones(query) {
```
psiemens: This method can be taken out -- you can just call `this.props.searchZones` directly

psiemens (changes requested):
This looks really clean. Nice use of the `ThemeManager`. Just one small fix before merging

psiemens (approved):

psiemens:
Looks good, merge away 👍

PR #757, opened 2018-01-19
514 implement twitter embed

lorenzodb1:
As we were talking about the need to resize the tweet, there is an issue related to the fact that the tweet style is enclosed into an <iframe>, which seems to be immutable. A potential solution could be to enclose it inside a <div> in which we can set the max-width as we prefer, but I had no success in adding it. Other than that, the embed seems to work fine.


.gitignore
```diff
@@ -16,3 +16,5 @@ htmlcov/
 dispatch/tests/media
 
 dispatch/apps/**/migrations/
+
+.idea/
```
psiemens: Can you add this to your global gitignore instead?

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  getTweet() {
+    const that = this
+    if (this.validateTweetURL(this.props.data.url)) {
+      fetchJsonp(TWITTER_API_URL + url.format({query: {url: this.props.data.url}}))
+      .then((response) => response.json())
+      .then(data => {
+        that.props.updateField('tweet', data.html)
+        this.props.stopEditing()
```
psiemens: I'm getting this error when I click "Insert":

screen shot 2018-01-19 at 4 36 06 pm

I suspect that it can be solved by calling `that.props.stopEditing()` instead of using `this`

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
```diff
@@ -0,0 +1,75 @@
...
+        <div className='o-embed--video__button__container'>
+          <Button onClick={() => this.getTweet()}>Insert</Button>
+        </div>
+      </div>
+    )
+  }
+
+  renderTweet() {
+    return (
+      <div>
```
psiemens: I don't think you need to have two divs here.

Also, can you add `1 rem` of padding around this div so that the tweet is better formatted?

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
```diff
@@ -0,0 +1,75 @@
...
+    )
+  }
+}
+
+export default {
+  type: 'tweet',
+  component: TwitterEmbedComponent,
+  defaultData: {
+    url: '',
+    id: '',
```
psiemens: I don't think we're using the `url` or `id` fields anymore.

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
```diff
@@ -0,0 +1,75 @@
...
+    tweet: ''
```
psiemens: This is nitpicky, but I think it might make more sense for this field to be named `html`.

dispatch/static/manager/webpack.config.js
```diff
@@ -50,5 +50,11 @@ module.exports = {
...
     new webpack.optimize.UglifyJsPlugin(),
     new ExtractTextPlugin('manager-' + version + '.css')
-  ]
+  ],
+  node: {
+    console: true,
+    fs: 'empty',
+    net: 'empty',
+    tls: 'empty'
+  }
```
psiemens, line 59: What's this for?

psiemens (changes requested):

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
```diff
@@ -0,0 +1,75 @@
...
+          <Button onClick={() => this.getTweet()}>Insert</Button>
+        </div>
+      </div>
+    )
+  }
+
+  renderTweet() {
+    return (
+      <div>
+        <div dangerouslySetInnerHTML={{__html: `${this.props.data.tweet}`}} />
```
JimFung: Don't know if this has been talked about in person at meetings or something, but I'd advise against using `dangerouslySetInnerHTML` as it does open you up to XSS attacks. This is especially dangerous when the data source is a third party (ie. twitter), and out of your control.

If you insist on using `dangerouslySetInnerHTML`, then you should first sanitize the html before injecting it.

dispatch/static/manager/webpack.config.js
lorenzodb1: I faced a problem in which the console would show "Error: Cannot find module "net"", and I solved it by adding that as explained here: https://github.com/request/request/issues/1529

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
lorenzodb1: We might need to see this on Wednesday as those div are actually not shown on the DOM, I have taken care of the other requests in the meantime

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
lorenzodb1: This was indeed discussed in person, but since I don't have too much experience with Javascript, HTML etc., I'm more than open to suggestions on how to replace that dangerouslySetInnerHTML!

JimFung: So for context, the danger here is there could be `<script>` tags in the tweet, which would allow a malicious actor to run any JS they want when we embed the tweet. I would just sanitize the html returned from the twitter api first before injecting it, its the simplest solution. 

The more involved solution would be to perform the tweet fetching on the server side and perform the sanitation there before passing clean html to the client. This would make it so users can't bypass the sanitation, which I suppose isn't really your responsibility; if they decide to bypass the sanitation, then they are responsible for the outcome. 

EDIT: There are many libraries for html sanitization btw, I'd just pick the most popular one and call it a day

dispatch/static/manager/src/js/vendor/dispatch-editor/embeds/TwitterEmbed.js
JimFung: anyone else think its weird the fetch isn't in a thunk? just me?

lorenzodb1: @JimFung would you mind expanding on that?

JimFung: @lorenzodb1 Peter and I talked about it when I visited a few days ago. I wouldn't worry about it, as it is now matches the design pattern that Peter wants for the embeds 👍

PR #763, opened 2018-02-04
Allow articles to have a featured video

### LINKED ISSUE #745: Allow articles to have a featured video
🚨  🚨  **PREREQUISITE: #746** 🚨  🚨  

We want to extend the Article model to support a `featured_video` field in addition to the existing `featured_image` field.

Backend requirements:
- Add a `featured_video` field to the [Article model](https://github.com/ubyssey/dispatch/blob/develop/dispatch/modules/content/models.py#L246)
  - Should be a foreign key to a new `VideoAttachment` model, similar to [ImageAttachment](https://github.com/ubyssey/dispatch/blob/develop/dispatch/modules/content/models.py#L480-L490)
- Extend the [ArticleSerializer](https://github.com/ubyssey/dispatch/blob/develop/dispatch/api/serializers.py#L388) to support the new `featured_video` field

Frontend requirements:
- Add an interface to the article editor that allows the user to set a featured video (in addition to a featured image)
  - This can be implemented as a new input called `VideoSelectInput`
  - Unlike the [ImageInput](https://github.com/ubyssey/dispatch/blob/develop/dispatch/static/manager/src/js/components/inputs/ImageInput.js), it doesn't need to open the image manager modal and can just be a simple `ItemSelectInput` (similar to [ArticleSelectInput](https://github.com/ubyssey/dispatch/blob/develop/dispatch/static/manager/src/js/components/inputs/selects/ArticleSelectInput.js))
  - After `VideoSelectInput` is created, it can be added to the [FeaturedImageTab](https://github.com/ubyssey/dispatch/blob/develop/dispatch/static/manager/src/js/components/Editor/tabs/FeaturedImageTab.js) component of the article editor

i-ron-y:
Close #745 

~There is a pre-existing bug that is affecting both image attachments and video attachments. Right now, you can create an article without an image or video attachment, and that's fine. But once you've added an image or video attachment to an article, the article must have an image or video attachment. You can change it to a different image or video, but you can't remove the attachment from the article entirely. The only way to do that is to delete the image or video from the database -- which is also buggy and doesn't work quite right.~

EDIT: Bug fixed.


psiemens:
Just to confirm -- is it just the coverage that's failing on this branch?

i-ron-y:
I think so! I tested it out and everything's working

psiemens (approved):
Let's merge this!

PR #765, opened 2018-02-08
Allow more than one nested widget in Zone Editor

i-ron-y:
Implemented as part of https://github.com/ubyssey/ubyssey.ca/issues/393

**Current bugs:**
- ~In the zone "FrontPage", when you select the "Facebook Video Big" widget, all the fields show [Object object]. You can change the value of the fields, but the field can't be empty - once it's empty, it shows [Object object] again. This seems to be related to the change in line 72 in dispatch/static/manager/src/js/components/ZoneEditor/index.js - when the change is removed, the Zone Editor page hangs if the currently-set widget has nested widget fields.~ FIXED

- In the zone "FrontPage", when the "Default Front Page" widget is selected (this is the widget for which we are going to have two nested widgets), set "Sidebar" to any widget, set "Sidebar2" to any widget, and then remove "Sidebar2". When you click "Update", this should work because "Sidebar2" isn't required, and it does work, but removing "Sidebar2" automatically removes "Sidebar" too for some reason. Now we're in a state that shouldn't exist, where the "Default Front Page" widget doesn't have a (required) "Sidebar" nested widget saved.

- Continuing on from where the second point leaves off, in that state where the "Default Front Page" widget is selected with neither of its nested widgets selected, if you try to set zone "FrontPage" to the "Facebook Video Big" widget, you get a "Zone could not be saved" error.


psiemens (approved):

psiemens:
@i-ron-y Let's go ahead and merge this, just because I think it will help Jamie and Rowan with some of the widget code they're working on.

PR #771, opened 2018-02-28
Attempt to fix nested widget saving bug

psiemens:


i-ron-y:
Working on tests. Is there a way to check coverage along the way, without having to push my commits through?

psiemens:
@i-ron-y You can run `dispatch-admin coverage`!

PR #775, opened 2018-03-20
774 photo metadata import

wmatous:
imports title, caption, tags, and author from image metadata (XMP not EXIF). currently prints out all fields to terminal with lines 466-479


dispatch/modules/content/models.py
```diff
@@ -401,14 +431,50 @@ def save(self, **kwargs):
...
+            print "title"
+            print self.title
+            print "caption"
+            print self.caption
+            print "tags"
+            for t in self.tags.all():
+                print t.name
+            print "author"
+            for a in self.authors.all():
+                print a
```
psiemens: Remove these print statements before merging

dispatch/modules/content/models.py
```diff
@@ -401,14 +431,50 @@ def save(self, **kwargs):
...
+            d = self.img.read()
+
+            xmp_start = d.find('<x:xmpmeta')
+            xmp_end = d.find('</x:xmpmeta')
+            #if the image contains XMP data, parse the relevant info
+            if xmp_start != -1:
+                xmp_str = d[xmp_start:xmp_end]
+                print xmp_str
+                self.title = self.find_xmp("title", xmp_str)
+                self.caption = self.find_xmp("description", xmp_str)
```
psiemens: These lines will set `title` and `caption` to empty strings (`''`) if they aren't set in the XML. It's probably a good idea to add a check to prevent this

psiemens (changes requested):

wmatous:
@psiemens looking gucci

PR #782, opened 2018-05-16
fix error in image upload

JamieRL:
Formdata was being passed to preparemultipartpayload from the image modal which was expecting an object.


Rowansdabomb (approved):
Looks good

PR #785, opened 2018-05-17
764 embeds freezing editor

JamieRL:
Fix editor freezing when opening embeds
simplify gallery select input
fix error when calling getEntityAt on null content block


dispatch/static/manager/src/js/vendor/dispatch-editor/components/Editor.js
```diff
@@ -147,6 +148,7 @@ class ContentEditor extends React.Component {
 
   stopEditingEntity() {
     this.setState({ readOnly: false })
+    this.focusEditor()
```
Rowansdabomb, line 150: redundent?

PR #788, opened 2018-05-18
Images multiple tags support

Rowansdabomb:


dispatch/static/manager/src/js/pages/Images/ImagesIndexPage.js
```diff
@@ -82,8 +83,8 @@ class ImagesPageComponent extends React.Component {
     this.props.clearSelectedImages()
   }
 
-  handleSearchImages(author, tags, query) {
-    this.props.searchImages(author, tags, query)
+  handleSearchImages(query) {
+    this.props.searchImages(null, null, query)
```
JamieRL: This implementation is good, storing everything in query is better than it was. We should probably change the definition of imageActions.search to just be search(query) so that we can remove these nulls

JamieRL (changes requested):
Looks good, just one bit of cleanup in the imageActions.search

dispatch/static/manager/src/js/pages/Images/ImagesIndexPage.js
JamieRL: NVM

The 'nulls' should get the current query.author and query.tags

JamieRL (approved):

PR #793, opened 2018-05-23
792 remove feature image

Rowansdabomb:


JamieRL (approved):
Looks good. Yayyyy more tests

PR #794, opened 2018-05-23
Fix create/save person

JamieRL:
Fix person create and save to pass raw object in now that the api handles making the form data


Rowansdabomb (approved):
Makes sense, looks good

PR #799, opened 2018-05-31
EXIF/XMP image metadata reading

wmatous:
automatic title, caption, author, tags (where available)


psiemens:
I think the problem is that the tests aren't passing, not that exempi isn't working. Running the image tests locally produces a bunch of errors for me, most of which look like this:

```bash
dispatch-admin test dispatch.tests.test_api_images
```

```
self.title = xmp.get_array_item(xmp.get_namespace_for_prefix('dc'), 'title', 1)
AttributeError: 'NoneType' object has no attribute 'get_array_item'
```

None of the test images have XMP data included, which is fine, but that's causing the `xmp` object to be `None`. That case should be handled, and it would also be a good idea to add a new test image (and accompanying test) that includes XMP data.

dispatch/modules/content/models.py
```diff
@@ -451,13 +459,76 @@ def save(self, **kwargs):
...
+            # Exif data is only read from files with TIFF or JPEG format, otherwise an error will occur
+            if self.get_extension() in Image.EXIF_FORMATS:
+                exiftags = image._getexif()
+
+                if exiftags is not None:
+                    # exiftags.get() returns None if no entry exists for a given key or if the entry is empty
+
+                    # exiftags object is a dictionary indexed by numeric tags corresponding to particular fields we may be
+                    # interested in. For a full list: https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html
+                    self.caption = exiftags.get(270)
```
psiemens: The `Image` model doesn't have a caption field. I would just leave this out and stick with `title`. That should fix your tests as well. 

Captions are only set when images are used as attachments in articles/galleries since the caption might depend on where the image is being used.

dispatch/modules/content/models.py
```diff
@@ -556,7 +557,9 @@ def save_thumbnail(self, image, size, name, label, file_type):
         default_storage.save(name, thumb_file)
 
     def save_tags(self, tag_ids):
-        self.tags.clear()
+
+        # self.tags.clear()
```
wmatous: @psiemens heres the thing i took out that was deleting all my tags, im not sure why my tags arent included in the validated data

dispatch/api/serializers.py
```diff
@@ -215,6 +216,7 @@ def update(self, instance, validated_data):
             instance.save_authors(authors)
 
         tag_ids = validated_data.get('tag_ids', False)
+
         if tag_ids != False:
```
wmatous: @psiemens heres the bit that calls save tags on "validated data"

PR #817, opened 2018-06-27
Fix publishable delete

JamieRL:
override destroy methods on article and page viewsets to delete the parent article/page


Rowansdabomb (approved):
Looks good

psiemens (approved):

PR #821, opened 2018-07-05
Fix event save

JamieRL:
Change the the save, publish, create, and, unpublish to no longer use form data as we process multipart requests in the API now.


Rowansdabomb (approved):
Looks ok to me

PR #824, opened 2018-07-09
490 section columns

JamieRL:


JamieRL:
Note, this PR also includes @Rowansdabomb 's menu bar redesign

dispatch/api/serializers.py
```diff
@@ -455,8 +456,90 @@ def load_data(self):
...
+        instance.slug = validated_data.get('slug', instance.slug)
+        instance.section_id = validated_data.get('section_id', instance.section_id)
+        instance.description = validated_data.get('description', instance.description)
+
+        # Save instance before processing/saving content in order to save associations to correct ID
+        instance.save()
+
+        # featured_image = validated_data.get('featured_image', False)
+        # if featured_image != False:
+        #     instance.save_featured_image(featured_image)
```
psiemens: Why is this commented out?

dispatch/api/validators.py
```diff
@@ -60,3 +60,20 @@ def AuthorValidator(data):
         if 'type' in author and not isinstance(author['type'], basestring):
             # If type is defined, it should be a string
             raise ValidationError('The author type must be a string.')
+
+class ColumnSlugValidator(object):
```
psiemens: Instead of making a separate slug validator, I think you can just use `SlugValidator`. IMO it doesn't matter if a column and a section share the same slug.

dispatch/modules/content/models.py
```diff
@@ -361,6 +362,43 @@ def get_absolute_url(self):
         """
         return "%s%s/%s/" % (settings.BASE_URL, self.section.slug, self.slug)
 
+    def get_column(self):
+        """
+        Returns the column set in the parent article
```
psiemens: Place this comment on single line

dispatch/modules/content/models.py
```diff
@@ -361,6 +362,43 @@ def get_absolute_url(self):
...
+        """
+        return self.parent.column
+
+    def save_column(self, column_id):
+        """
+        Save the column to the parent article
```
psiemens: Place this comment on single line

psiemens (dismissed):

dispatch/api/validators.py
JamieRL: So here's where I ran into some confusion with how we structure the urls. I was deciding between whether I should do `baseurl/section_slug/column_slug/` or `baseurl/column_slug/` . I ended up choosing `baseurl/column_slug` because I thought it would be easier to enforce columns and sections not having the same slugs over columns and articles not having the same slug

dispatch/api/serializers.py
JamieRL: I wasn't quite sure at the time whether we would end up using featured images for columns and forgot to resolve it. What do you think? For the front end I ended up copying the section template fairly closely

dispatch/api/validators.py
```diff
@@ -64,6 +64,23 @@ def AuthorValidator(data):
             # If type is defined, it should be a string
             raise ValidationError('The author type must be a string.')
 
+class SubsectionSlugValidator(object):
+    def set_context(self, serializer_field):
```
Rowansdabomb: lets move this logic to the slugvalidator

dispatch/api/serializers.py
```diff
@@ -4,15 +4,16 @@
...
+    VideoAttachment, Poll, PollAnswer, PollVote, Subsection)
 from dispatch.modules.auth.models import Person, User, Invite
 from dispatch.admin.registration import send_invitation
 from dispatch.theme.exceptions import WidgetNotFound, InvalidField
 
 from dispatch.api.mixins import DispatchModelSerializer, DispatchPublishableSerializer
 from dispatch.api.validators import (
     FilenameValidator, ImageGalleryValidator, PasswordValidator,
-    SlugValidator, AuthorValidator, TemplateValidator)
+    SlugValidator, AuthorValidator, TemplateValidator, SubsectionSlugValidator)
```
Rowansdabomb: remove subsectionSlugValidator

dispatch/api/views.py
```diff
@@ -31,7 +31,7 @@
     FileSerializer, IssueSerializer, ImageGallerySerializer, TagSerializer,
     TopicSerializer, PersonSerializer, UserSerializer, IntegrationSerializer,
     ZoneSerializer, WidgetSerializer, TemplateSerializer, VideoSerializer,
-    PollSerializer, PollVoteSerializer, InviteSerializer)
+    PollSerializer, PollVoteSerializer, InviteSerializer, SubsectionSerializer)
```
Rowansdabomb, line 34: remove SubsectionSlugValidator

dispatch/tests/test_api_columns.py
```diff
@@ -0,0 +1,268 @@
+from django.core.urlresolvers import reverse
+
+from rest_framework import status
+
+from dispatch.tests.cases import DispatchAPITestCase
+from dispatch.tests.helpers import DispatchTestHelpers
+from dispatch.models import Subsection, Article, Person, Section
+
+
```
Rowansdabomb: extra space

Rowansdabomb (changes requested):

Rowansdabomb (approved):

PR #825, opened 2018-07-10
492 push notifications

JamieRL:
Dispatch notification branch


psiemens:
I got this working locally, and for the most part everything is super slick. I'm amazed by how simple the whole thing is -- I thought it would require more browser/OS integration on our part. 

Before I get into a more detailed code review, I wanted to bring up a few concerns and propose some changes. My two main concerns with the current solution are spam for users and lack of control for editors/admins. Creating a notification automatically for every article is slick/effortless, but I worry that users might complain about a bunch of notifications coming in at the same time every day. Also, I'm not sure that _every_ article needs to have a notification. We publish a lot of articles during the school year and I could see notifications getting annoying and even start to lose their meaning if every article is pushed to the user.

I noticed that you added some comments about a breaking news option, which I assume delivers a notification immediately rather than waiting until the next cron interval. What if we took this a step further and allowed editors to schedule every push notification? My rough ideas:

* Editors can choose which articles will be pushed to the user
* From the article page, they can check a box that says "send push notification"
* Under that checkbox, their could be an option to "send immediately upon publish" or at a specific date/time

I think this would allow us to better distribute notifications throughout the day and would make each individual notification a bit more meaningful to the user. Thoughts? @Rowansdabomb @JamieRL

JamieRL:
I think what we have currently wouldn't spam users (assuming we scheduled the cron task properly), however the queue would get backlogged if a lot of articles were published in a short time. 

I'm all for giving the editors more control and letting them choose which articles to push a notification with sounds like a good idea to me. 

Do we make fb/twitter post at the time each new article is published? If so I think we should treat push notifications a little different (make sure we space them out) since its a bit more intrusive for the user

JamieRL:
Also, I'm working on the breaking news thing separately and am almost done so we could integrate it with the notifications

psiemens:
> Do we make fb/twitter post at the time each new article is published? If so I think we should treat push notifications a little different (make sure we space them out) since its a bit more intrusive for the user

We do, but the editors use Hootsuite to schedule all the social media posts. Dispatch doesn't handle that...yet.

JamieRL:
So I added a "schedule notification" under the delivery tab in the article editor that lets you pick when a notification would be sent out. we just have to figure out how to get the cron task to send the notification at the specified time

JamieRL:
one way could be to just get the cron job to every 15-30 minutes check the notifications, see if there is a notification who's scheduled time has passed, and if there is, push that notification and delete it. (this would also limit the number of notifications being pushed out to 1 per 15-30 minutes)

JamieRL:
So I set it up like this `Notification.objects.filter(scheduled_push_time__lte=timezone.now()).order_by('scheduled_push_time').first()` to get the notification that should be pushed out first. Now if we can schedule our cron task to periodically check then this should work

JamieRL:
Can't quite get it to pass travis, its saying that pywebpush requries a difference version of requests

dispatch/api/serializers.py
```diff
@@ -845,3 +849,46 @@ def update(self, instance, validated_data, is_new=False):
...
+        # Create new Subscription instance
+        instance = Subscription()
+
+        instance.endpoint = validated_data.get('endpoint', instance.endpoint)
+        instance.auth = validated_data.get('auth', instance.auth)
+        instance.p256dh = validated_data.get('p256dh', instance.p256dh)
+
+        instance.save()
+
+        return instance
```
psiemens: This looks like the default functionality for a `create` method -- I'm not sure if you even need this.

dispatch/api/views.py
```diff
@@ -93,6 +99,28 @@ def get_queryset(self):
...
+
+            relation, created = ArticleRelation.objects.get_or_create(parent=article, article=related_article)
+
+            if not created:
+                relation.count += 1
+                relation.save()
+        else:
+            return Response('')
+
+        return Response('')
```
psiemens: Is this method supposed to be part of this PR?

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
         return Response(serializer.data)
 
+class NotificationsViewSet(DispatchModelViewSet):
+    """Viewset for the Poll model views."""
+
+    model = Notification
+    serializer_class = NotificationSerializer
+    permission_classes = (IsAuthenticated,)
+
+
```
psiemens: Extra line break

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+    def get_queryset(self):
+        queryset = Notification.objects.all().order_by('created_at')
```
psiemens: Just return this on one line

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+        return queryset
+
+    @detail_route(permission_classes=[AllowAny], methods=['post', 'patch'],)
+    def subscribe(self, request, pk=None):
+        data = {
+            'endpoint': request.data['endpoint'],
+            'auth':request.data['keys']['auth'],
+            'p256dh':request.data['keys']['p256dh'],
+        }
```
psiemens: Add line break after this

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+
+    def get_queryset(self):
+        queryset = Notification.objects.all().order_by('created_at')
+        return queryset
+
+    @detail_route(permission_classes=[AllowAny], methods=['post', 'patch'],)
+    def subscribe(self, request, pk=None):
+        data = {
+            'endpoint': request.data['endpoint'],
+            'auth':request.data['keys']['auth'],
```
psiemens: Add space after key `'auth': request.data['keys']['auth'],` (same below)

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+            'p256dh':request.data['keys']['p256dh'],
+        }
+        try:
+            serializer = SubscriptionSerializer(data=data)
+            serializer.is_valid(raise_exception=True)
+            serializer.save()
+            return Response(serializer.data)
+        except:
+            print('User already subscribed')
```
psiemens: Can this return an error instead?

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+            try:
+                webpush(
+                    subscription_info={
+                        "endpoint": sub.endpoint,
+                        "keys": {
+                            "p256dh": sub.p256dh,
+                            "auth": sub.auth
+                        }},
+                    data=json.dumps(data),
+                    vapid_private_key="Mp2OSApC5ZQ11iHtKfTfAWycrr-YYl9yphpkeqKIy9E",
```
psiemens: This should be stored in `settings.py`

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+
+        return Response(serializer.data)
+
+    def push_notifications(self, article):
+        # grab each endpoint from list in database and make a push
+        data={
+            'headline': article.headline,
+            'url': article.get_absolute_url(),
+            'snippet': article.snippet,
+            'image': article.featured_image.image.get_thumbnail_url()
```
psiemens: Check that featured_image is set, exclude `image` if it is missing

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+                    subscription_info={
+                        "endpoint": sub.endpoint,
+                        "keys": {
+                            "p256dh": sub.p256dh,
+                            "auth": sub.auth
+                        }},
+                    data=json.dumps(data),
+                    vapid_private_key="Mp2OSApC5ZQ11iHtKfTfAWycrr-YYl9yphpkeqKIy9E",
+                    vapid_claims={
+                            "sub": "mailto:YourNameHere@example.org===",
```
psiemens: Is this field required?

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+        notification = Notification.objects.filter(scheduled_push_time__lte=timezone.now()).order_by('scheduled_push_time').first()
+
+        if notification is not None:
+            self.push_notifications(notification.article)
+            notification.delete()
+        return Response({'detail': 'success'})
+
+
+
+
```
psiemens: Remove extra lines

dispatch/modules/content/models.py
```diff
@@ -323,8 +325,11 @@ class Article(Publishable, AuthorMixin):
     def title(self):
         return self.headline
 
-    def get_related(self):
-        return Article.objects.exclude(pk=self.id).filter(section=self.section,is_published=True).order_by('-published_at')[:5]
+    def get_related(self, n=5):
+        """
+        Returns n related articles.
+        """
```
psiemens: Put this all on one line: `"""Returns n related articles."""`

dispatch/modules/content/models.py
```diff
@@ -323,8 +325,11 @@ class Article(Publishable, AuthorMixin):
...
+        return Article.objects.exclude(pk=self.id).filter(section=self.section,is_published=True).order_by('-published_at')[:n]
```
psiemens: This is a bit cleaner:

```py
return Article.objects \
  .exclude(pk=self.id) \
  .filter(section=self.section, is_published=True) \
  .order_by('-published_at')[:n]

File: dispatch/modules/content/models.py [hunk 14]
```diff
@@ -361,6 +366,11 @@ def get_absolute_url(self):
         """
         return "%s%s/%s/" % (settings.BASE_URL, self.section.slug, self.slug)
 
+class ArticleRelation(Model):
+    parent = ForeignKey('Article', related_name='parent_article')
+    article = ForeignKey('Article', related_name='article')
+    count = PositiveIntegerField(default=1)
+
```
psiemens: I think this should be part of a separate PR since it's not related to notifications. Correct me if I'm wrong

dispatch/static/manager/src/js/components/Editor/tabs/DeliveryTab.js
```diff
@@ -76,6 +81,12 @@ export default function DeliveryTab(props) {
         {warningMessage}
       </FormInput>
 
+      <FormInput label='Schedule Notification'>
+        <DateTimeInput
+          value={props.scheduled_notification}
+          onChange={dt => updateScheduledNotification(props.update, dt)} />
```
psiemens: I would just do this:

```js
dt => props.update('scheduled_notification', dateObjToAPIString(dt))
```

psiemens: Also, can you add a way to remove (cancel) a notification once it has been created?

dispatch/api/views.py
```diff
@@ -287,6 +315,74 @@ def vote(self, request, pk=None):
...
+                    vapid_claims={
+                            "sub": "mailto:YourNameHere@example.org===",
+                        }
+                )
+            except WebPushException as ex:
+                sub.delete()
+
+    @list_route(permission_classes=[IsAuthenticated], methods=['post'],)
+    def push(self, request, pk=None):
+        notification = Notification.objects.filter(scheduled_push_time__lte=timezone.now()).order_by('scheduled_push_time').first()
```
psiemens:
```py
notification = Notification.objects \
  .filter(scheduled_push_time__lte=timezone.now()) \
  .order_by('scheduled_push_time') \
  .first()

File: dispatch/api/serializers.py [hunk 17]
```diff
@@ -845,3 +849,46 @@ def update(self, instance, validated_data, is_new=False):
             instance.save_answers(answers, is_new)
 
         return instance
+
+class SubscriptionSerializer(DispatchModelSerializer):
+    """Serializes the Subscription model."""
+
+    endpoint = serializers.CharField(required=True, write_only=True)
+    auth = serializers.CharField(required=True, write_only=True)
+    p256dh = serializers.CharField(required=True, write_only=True)
```
psiemens, line 939: Should the auth fields be encrypted?

psiemens (changes requested):

dispatch/modules/content/models.py
JamieRL: yeah, this branch was off of the new suggested article branch that rowan started I think

JamieRL: I removed the article relation stuff

dispatch/api/views.py
JamieRL: removed, along with other suggested article stuff

JamieRL:
Need to add page for managing notifications

look into webpushexception response

dispatch/api/serializers.py
```diff
@@ -3,12 +3,13 @@
 from rest_framework.validators import UniqueValidator
 
 from dispatch.modules.content.models import (
-    Article, Image, ImageAttachment, ImageGallery, Issue,
-    File, Page, Author, Section, Tag, Topic, Video, VideoAttachment, Poll, PollAnswer, PollVote)
+    Article, Image, ImageAttachment, ImageGallery, Issue, Subscription,
+    File, Page, Author, Section, Tag, Topic, Video, VideoAttachment, Poll, PollAnswer, PollVote, Notification, SubscriptionCount)
```
Rowansdabomb: lets more evenly distribute these two lines

dispatch/api/serializers.py
```diff
@@ -3,12 +3,13 @@
...
 from dispatch.modules.auth.models import Person, User, Invite
 from dispatch.admin.registration import send_invitation
 from dispatch.theme.exceptions import WidgetNotFound, InvalidField
 
+
```
Rowansdabomb: extra space

dispatch/api/views.py
```diff
@@ -31,14 +38,17 @@
...
     TopicSerializer, PersonSerializer, UserSerializer, IntegrationSerializer,
     ZoneSerializer, WidgetSerializer, TemplateSerializer, VideoSerializer,
-    PollSerializer, PollVoteSerializer, InviteSerializer)
+    PollSerializer, PollVoteSerializer, NotificationSerializer, InviteSerializer,
+    SubscriptionSerializer, SubscriptionCountSerializer)
 from dispatch.api.exceptions import (
     ProtectedResourceError, BadCredentials, PollClosed, InvalidPoll,
-    UnpermittedActionError)
+    UnpermittedActionError, AlreadySubscribed)
+
```
Rowansdabomb: extra white space

dispatch/api/views.py
```diff
@@ -31,14 +38,17 @@
...
 
 from dispatch.theme import ThemeManager
 from dispatch.theme.exceptions import ZoneNotFound, TemplateNotFound
 
+
```
Rowansdabomb: extra space

dispatch/api/views.py
```diff
@@ -387,6 +397,79 @@ def vote(self, request, pk=None):
...
         return Response(serializer.data)
 
+class SubscriptionCountViewSet(DispatchModelViewSet):
+    """Viewset for the SubscriptionCount model views."""
+
+    model = SubscriptionCount
+    serializer_class = SubscriptionCountSerializer
+    queryset = SubscriptionCount.objects.filter(date__gte=timezone.now() - datetime.timedelta(days=90)).order_by('-date')
+
+
```
Rowansdabomb: extra space

dispatch/api/views.py
```diff
@@ -387,6 +397,79 @@ def vote(self, request, pk=None):
...
+    @list_route(permission_classes=[AllowAny], methods=['post'],)
+    def push(self, request):
+        notification = Notification.objects \
+        .filter(scheduled_push_time__lte=timezone.now()) \
+        .order_by('scheduled_push_time') \
+        .first()
+
+        if notification is not None:
+            article = Article.objects.filter(parent__id=notification.article.parent_id, is_published=True).first()
```
Rowansdabomb: lets change this from filter to get

dispatch/default_settings.py
```diff
@@ -90,4 +90,5 @@
 PHONENUMBER_DB_FORMAT = 'NATIONAL'
 PHONENUMBER_DEFAULT_REGION = 'CA'
 
+NOTIFICATION_KEY = "Mp2OSApC5ZQ11iHtKfTfAWycrr-YYl9yphpkeqKIy9E"
```
Rowansdabomb: make this secret in settings-prd.py

dispatch/migrations/0018_subscriptioncount.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
Rowansdabomb: squash 0017 and 0018 migrations into 0017_notifications

dispatch/static/manager/src/js/components/ItemEditor/index.js
```diff
@@ -89,7 +89,7 @@ class ItemEditor extends React.Component {
             deleteListItem={() => this.props.deleteListItem(this.props.token, this.props.itemId, this.props.afterDelete)}
             goBack={this.props.goBack}
             extraButton={this.props.extraButton} />
-          <div className='u-container u-container--padded u-container--vscroll'>
+          <div className={'u-container u-container--padded '+(this.props.showOverflow ? '' : 'u-container--vscroll')}>
```
Rowansdabomb: remove the + symbol

dispatch/static/manager/src/js/pages/Notifications/NotificationIndexPage.js
```diff
@@ -0,0 +1,131 @@
...
+  renderChart() {
+    let dates = []
+    let counts = []
+
+    for (var subscriptionCount in this.props.subscriptionCounts) {
+      dates.push(humanizeDatetime(this.props.subscriptionCounts[subscriptionCount].date))
+      counts.push(this.props.subscriptionCounts[subscriptionCount].count)
+    }
+
+    let chartData = {
```
Rowansdabomb: make chartData const

dispatch/static/manager/src/styles/components/notifications.scss
```diff
@@ -0,0 +1,26 @@
+
+
```
Rowansdabomb: extra space

dispatch/static/manager/src/styles/components/notifications.scss
```diff
@@ -0,0 +1,26 @@
...
+.c-subscription-chart-container {
```
Rowansdabomb: comment sass

Rowansdabomb (changes requested):

dispatch/static/manager/src/js/components/ItemEditor/index.js
JamieRL: The + is actually necessary otherwise you get an error saying: "u-container u-container--padded is not a function"

PR #826, opened 2018-07-12
818 breaking news

JamieRL:
- Added a breaking news toggle in the article editor top bar
 - Added is_breaking field to the article moddle
 - Added a currently_breaking field to the serializer that checks if the article is: published, breaking, published less than 2 hours ago


dispatch/api/serializers.py
```diff
@@ -522,6 +524,9 @@ class Meta:
             'section_id',
             'published_at',
             'is_published',
+            'is_breaking',
+            'breaking_timeout',
+            'currently_breaking',
```
JamieRL: I'm thinking that since I added the currently_breaking field to the serializer that the 'is_breaking' field may not even be needed anymore

Rowansdabomb (approved):
looks just swell

PR #827, opened 2018-07-20
513 timeline

Rowansdabomb:


dispatch/api/serializers.py
```diff
@@ -560,6 +560,8 @@ def update(self, instance, validated_data):
         instance.template = validated_data.get('template_id', instance.template)
         instance.template_data = validated_data.get('template_data', instance.template_data)
 
+        if instance.template == 'timeline':
+            TimelineValidator(validated_data.get('template_data', None))
```
psiemens, line 564: Keep this for now, but can you add a TODO?

```py
# TODO: create abstract template validator

# Example:
TemplateValidator(instance.template)
```

I think a more general approach would be to specify required fields and data types in the template definition and then check those in a generic validator. This way any template can be validated in the future.

dispatch/static/manager/src/js/components/inputs/DateTimeInput.js
```diff
@@ -11,6 +11,10 @@ const MAX_DATE = new Date(TODAY.getFullYear()+1, 12, 31)
 function ensureDate(date) {
   let ret = date
   if (!(date instanceof Date)) {
+    // // console.log(Date.parse(date.slice(0, -1)))
+    // let index = date.indexOf('T')
+    // date = date.slice(0, index)
+    // console.log(Date.parse(date))
```
psiemens: Remove old comments

dispatch/theme/fields.py
```diff
@@ -333,3 +333,15 @@ class PollField(ModelField):
...
+
+class InstructionField(Field):
+    type = 'instruction'
+
+    def __init__(self, label, options=[], required=False):
+        self.options = options
+        self.valid_options = set(option[0] for option in self.options)
+
+        # if required and not self.options:
+        #     raise InvalidField('Empty select fields cannot be required fields')
```
psiemens: Remove old comment

psiemens (changes requested):

PR #831, opened 2018-07-24
518 image wrap

Rowansdabomb:


dispatch/migrations/0014_auto_20180724_0922.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11 on 2018-07-24 16:22
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
```
JamieRL: We should squash these migrations and only include the relevant ones. 
Run these commands and re-commit with only 1 migration named 0013_image_style.py
`python manage.py migrate dispatch 0012`
(delete all the new migration files (0013 to 0016)
`python manage.py makemigrations`

JamieRL (changes requested):

PR #837, opened 2018-07-30
Develop beta

Rowansdabomb:


dispatch/static/manager/src/js/util/helpers.js
```diff
@@ -11,6 +11,15 @@ export function humanizeDatetime(timestamp, isDateTime) {
     moment(timestamp).format('ll')   // April 18, 2017
 }
 
+export function pythonifyDateTime(timestamp) {
```
JamieRL: Does the existing helper function "dateObjToAPIString" work for this?

dispatch/migrations/0015_image_wrap.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11 on 2018-07-27 00:25
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
```
JamieRL, line 6: We'll have to change this migration again now that we unmerged metadata

JamieRL (changes requested):
Just a few things we'll have to change before merging.

JamieRL (approved):

PR #853, opened 2018-08-08
Make publishable save more robust

JamieRL:
add integrity check for if two articles are set to head, make transaction atomic


JamieRL:
Unsure of what to do with the tests for the moment. Gonna hold off on this one for a bit

PR #861, opened 2018-08-29
Fix subsection page

JamieRL:
- fix invocation of save_subsection in article serializer to be able to remove a subsection
 - add ordering by published_at to get_published_articles in subsection model
 - add missing prop to basicfieldstab in articlesidebar.js so that selected subsection shows up properly


Rowansdabomb (approved):
checks out

PR #864, opened 2018-09-05
Fix new article page from dashboard

JamieRL:
check if template exists before accessing. This fixes refreshing the new article page as well as navigating to new article page from the dashboard


Rowansdabomb (approved):
Looks good

PR #874, opened 2018-09-27
Update form styles

psiemens:


dispatch/static/manager/src/js/components/PersonEditor/PersonForm.js
```diff
@@ -32,53 +35,54 @@ export default class PersonForm extends React.Component {
```
Rowansdabomb, line 35: Move constructor to top of Component

dispatch/static/manager/src/js/components/PollEditor/PollForm.js
```diff
@@ -124,8 +127,9 @@ export default class PollForm extends React.Component {
     return (
       <div className={'c-poll-form-container'}>
         <div className={'c-equal-width'}>
-          <form onSubmit={e => e.preventDefault()}>
-            <FormInput
```
Rowansdabomb, line 128: This is cool

Rowansdabomb (commented):
Looks Good. There is just one constructor that should be moved to the top of the component for consistency.

dispatch/static/manager/src/js/components/PersonEditor/PersonForm.js
psiemens: nice catch 👍

PR #876, opened 2018-10-03
Fix topic slug migration

psiemens:
* Add a separate migration for the `Topic.slug` field
* Auto-generate slugs using `Topic.name`
* Fix migration tree so that it matches the old structure


Rowansdabomb (approved):
Looks good

PR #877, opened 2018-10-04
873 release podcasts

Rowansdabomb:


jumbosushi:
@Rowansdabomb hmm im seeing 404 page for `/podcasts`. Am I missing a step?

dispatch/modules/podcasts/models.py
```diff
@@ -84,7 +87,13 @@ def __init__(self, *args, **kwargs):
     file = FileField(upload_to='podcasts/')
 
     def get_absolute_url(self):
-        return self.file.url
+        return "%spodcast/%s#%s" % (settings.BASE_URL, self.podcast.slug, self.get_slug_from_title())
+
+    def get_file_url(self):
+        return "%s%s/" % (settings.MEDIA_URL, self.file)
+
+    def get_slug_from_title(self):
```
psiemens: I think it might be better to add an actual `slug` field to the model, even if it's auto-generated from the title on the first save (see the `Topic` mode). Once a URL is published (i.e. on Facebook), it's important to keep the URL the same to prevent dead links. In this case, the URL could change if somebody updates the title.

Alternatively, could you just use the episode number?

dispatch/modules/podcasts/models.py
```diff
@@ -84,7 +87,13 @@ def __init__(self, *args, **kwargs):
     file = FileField(upload_to='podcasts/')
 
     def get_absolute_url(self):
-        return self.file.url
+        return "%spodcast/%s#%s" % (settings.BASE_URL, self.podcast.slug, self.get_slug_from_title())
```
psiemens: I think it would be best to remove this method and instead implement it as a helper in `ubyssey.ca`. Different dispatch themes will use different paths for podcasts, but this enforces a `podcasts/my-podcast/my-episode` pattern.

Same for `Podcast.get_absolute_url`

dispatch/core/storage.py
```diff
@@ -9,7 +9,17 @@ def generate_signed_url(path, content_type, expires_after_seconds=60):
 
     obj = os.path.join(settings.GS_LOCATION, path)
 
-    client = storage.Client()
+    if settings.LOCAL_DEV:
+        dirname = os.path.dirname
+        ##############################################################################
+        # Ensure you have a valid service-account json key file and set its path below
+        ##############################################################################
+        key_file = os.path.join('/', 'ubyssey.ca', 'ubyssey.ca', 'gcs-local.json')
```
psiemens: This should be a configurable setting (e.g. `settings.GCS_CREDENTIALS_FILE`) instead of a hard-coded variable. The above will only work inside the `ubyssey.ca` repo

dispatch/modules/podcasts/models.py
jumbosushi: sidenote django have `slugify` method
```
>>> from django.template.defaultfilters import slugify
>>> slugify("b b b b")
u'b-b-b-b'
>>>
```

Rowansdabomb: I think ill use the episode number to avoid the migration

dispatch/modules/podcasts/models.py
Rowansdabomb: The episodes do not have their own pages. So how would if I move the `get_absolute_url` function to `helpers.py`, how would i access it from the template?

Rowansdabomb: I will just zip the episode data together with the url in the `main.py`

Rowansdabomb:
@psiemens and @jumbosushi do these changes look ok?

psiemens (approved):

psiemens:
@Rowansdabomb LGTM

jumbosushi (approved):
:100:  (assuming `LOCAL_DEV` is from [the other pr](https://github.com/ubyssey/ubyssey.ca/pull/571/files) )

PR #886, opened 2018-10-15
this should make the 0020_topic_slug migration reversible

Rowansdabomb:


dispatch/migrations/0021_topic_slug_unique.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
psiemens: Was it necessary to create a whole new migration file?

Rowansdabomb: When applying the reverse_set_defaults function, I ran into a duplicate key error. So the second migration file is just to ensure that before reverting migration 0020, the fields do not have to be unique. Of course as I write this I realize I set the slugs to None, so you are probably right, this 0021 isn't necessary.

psiemens (approved):

psiemens:
Looks good!

jumbosushi:
Looks awesome :100:

PR #892, opened 2018-10-31
Prevent duplicate head, is_published fields with database constraints

### LINKED ISSUE #888: Prevent duplicate head entries for articles
It looks like there's still a race condition that causes multiple articles with `head=True` to exist in the database. We previously tried to prevent this with the `is_conflicting_head` method, but I don't know if that fixes the problem. I'm actually worried that since `article.save()` is grouped into an atomic transaction, `is_conflicting_head` won't help us in the case where multiple requests are saving the same article. 

A proposed fix:
- Use a raw SQL query to perform the version update
  - This is similar to using `@transaction.atomic`
- Add a MySQL UNIQUE constraint for `(parent_id, head)` 
  - This will prevent the database from ever saving two articles with `head=1`

psiemens:
* Added database constraints using `unique_together` to prevent multiple `head` or `is_published` fields from being `True` for the same article
* Added custom `NullBooleanField` to use in `ArticleSerializer` and `PageSerializer`

Closes #888


dispatch/modules/content/models.py
```diff
@@ -184,19 +184,17 @@ def save(self, revision=True, *args, **kwargs):
...
                 # Clear the instance id to force Django to save a new instance.
                 # Both fields (pk, id) required for this to work -- something to do with model inheritance
                 self.pk = None
                 self.id = None
 
                 # New version is unpublished by default
-                self.is_published = False
-
-        # Raise integrity error if instance with given slug already exists.
-        if type(self).objects.filter(slug=self.slug).exclude(parent=self.parent).exists():
```
Rowansdabomb, line 198: So using the meta option `unique_together` achieves the same result?

Rowansdabomb (approved):
LGTM

Issue #897, opened 2018-11-16
Order images from newest


jumbosushi:
Right now images are shown from oldest to newest, which is really inconvenient for editors. This should be from newest to oldest

![highlight](https://user-images.githubusercontent.com/9669739/48604538-03b92480-e92f-11e8-8e4a-7a23bedd959c.png)

PR #898, opened 2018-11-16
Show images from newest

jumbosushi:
Old `/images` pages shows the images from newest to oldest, but this pr changes to the other way as described in #897 
## Before
image


Rowansdabomb (approved):
LGTM

PR #901, opened 2018-11-16
Add micro improvement to Aricle queryset

jumbosushi:
`Article` model have new ForeignKey relationships, so the queryset should reflect this by adding these new fields to `select_related` for micro speed boost

### DIFF
--- a/dispatch/api/views.py
+++ b/dispatch/api/views.py
@@ -82,7 +82,7 @@ class ArticleViewSet(DispatchModelViewSet, DispatchPublishableMixin):
 
         # Optimize queries by prefetching related data
         queryset = queryset \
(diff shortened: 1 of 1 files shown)


dispatch/api/views.py
```diff
@@ -82,7 +82,7 @@ def get_queryset(self):
 
         # Optimize queries by prefetching related data
         queryset = queryset \
-            .select_related('featured_image', 'topic', 'section') \
+            .select_related('featured_image', 'featured_video', 'topic', 'section', 'subsection') \
```
Rowansdabomb, line 85: I do not think `featured_video` is not ready for release yet

Rowansdabomb (changes requested):
just a little thing

PR #909, opened 2019-02-25
add articleListSerialzier

Rowansdabomb:
## What 
Speeds up the articles listed on the dispatch/admin/articles/ page.
image

## How
Adds an `ArticleListSerializer` that requests only a portion of the articles information instead of the entire article content. 
## Side effects
The article content must load once the article is clicked, but this is only done for 1 article instead of thousands. 

Hard to say how much affect this will have in production. A lot of nested serializers are removed, so it is less likely to cause N+1 sql queries. 

## Before
image

## After
image

### DIFF
--- a/dispatch/api/views.py
+++ b/dispatch/api/views.py
@@ -28,7 +28,7 @@ from dispatch.admin.registration import reset_password
 
 from dispatch.api.mixins import DispatchModelViewSet, DispatchPublishableMixin
 from dispatch.api.serializers import (
-    ArticleSerializer, PageSerializer, SectionSerializer, ImageSerializer,
+    ArticleListSerializer, ArticleSerializer, PageSerializer, SectionSerializer, ImageSerializer,
     FileSerializer, IssueSerializer, ImageGallerySerializer, TagSerializer,
     TopicSerializer, PersonSerializer, UserSerializer, IntegrationSerializer,
     ZoneSerializer, WidgetSerializer, TemplateSerializer, VideoSerializer,
@@ -72,6 +72,13 @@ class ArticleViewSet(DispatchModelViewSet, DispatchPublishableMixin):
     model = Article
     serializer_class = ArticleSerializer
     lookup_field = 'parent_id'
+
+    def get_serializer_class(self):
+        try:
+            self.request.query_params['limit']
+        except:
+            return ArticleSerializer
+        return ArticleListSerializer
 
     def get_queryset(self):
         """Optionally restricts the returned articles by filtering against a `topic`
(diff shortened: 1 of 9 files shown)


dispatch/api/views.py
```diff
@@ -73,6 +73,13 @@ class ArticleViewSet(DispatchModelViewSet, DispatchPublishableMixin):
     serializer_class = ArticleSerializer
     lookup_field = 'parent_id'
 
+    def get_serializer_class(self):
```
Rowansdabomb: open to suggestions on a better way to do this

dispatch/static/manager/src/js/actions/ArticlesActions.js
```diff
@@ -42,7 +42,12 @@ class ArticlesActions extends PublishableActions {
   }
 
   fromRemote(data) {
-    data.content = fromJSON(data.content)
+    try {
```
Rowansdabomb: This could probably be cleaner?

dispatch/api/views.py
psiemens: I'm just worried about cases where an API consumer would want to fetch a list of articles with the full content included

dispatch/static/manager/src/js/actions/ArticlesActions.js
psiemens: What does `fromJSON(data.content)` throw if `data.content` is null? I would maybe refactor `fromJSON` to accept (and ignore) null values and then just call it normally: 

```js
function fromJSON(data) {
  if (!data) {
    return data
  }

  // convert content to JSON
}

**Review by psiemens — COMMENTED** [review]
Great optimization! The pre-loading of articles was smooth at first, but it definitely shouldn't be something that kills performance.

I think there might be a way to generalize this behaviour for other endpoints without too much code duplication. It would be cool to allow an optional query param `exclude` that lists the fields to be excluded from the response: `api/articles/?limit=10&exclude=content,featured_image`

You could then modify `DispatchModelSerializer` to dynamically pop the excluded fields from the serializer. We're already doing this inside `hide_authenticated_fields`:
https://github.com/ubyssey/dispatch/blob/develop/dispatch/api/mixins.py#L86-L104

Query params can be accessed from inside `DispatchModelSerializer` like so:

```py
request = self.context.get('request')
if request:
  exclude = request.query_params.get('exclude', '')
  excluded_fields = exclude.split(',')
```

dispatch/static/manager/src/js/actions/ArticlesActions.js
Rowansdabomb: I like this idea

jumbosushi (approved):
Looks really great I'm so excited for this :tada:

Issue #912, opened 2019-03-07
Exlcude query breaks on non article index pages


Rowansdabomb:
The exclude query used to optimise the articles index page is breaking other pages that use the  `ItemIndexPage` component.

Rowansdabomb:
Fixed in #913

PR #913, opened 2019-03-07
conditionally add exclude query only when passed from parent

Rowansdabomb:
#912 

This only adds the exclude query when passed by the parent component. Thus it can be handled on a case by case basis in the future.


dispatch/static/manager/src/js/pages/ItemIndexPage.js
```diff
@@ -15,10 +15,15 @@ const DEFAULT_LIMIT = 15
 export default class ListItemsPageComponent extends React.Component {
 
   getQuery() {
+    console.log(this.props.exclude)
```
jumbosushi: Let's remove this `console.log`

Rowansdabomb: whoops, thanks for catching this

Issue #910, opened 2019-03-07
Add section to articles page


Rowansdabomb:
Currently the articles page on dispatch looks like this:
![image](https://user-images.githubusercontent.com/19273530/53990119-6734b100-40dc-11e9-9f2f-d0a5540747a5.png)

It would be convenient to list the section of the article in addition to is headline, authors, etc

Rowansdabomb:
Fixed in commit d8f722c40182590647444fb3f0e7393683dce4cf

PR #914, opened 2019-03-07
add sections column to the articles index page

Rowansdabomb:
#910 

There is now a sections column on the `ArticleIndexPage` component.

image


jumbosushi (approved):
Love it :raised_hands:

PR #916, opened 2019-03-16
Add 'overflow: auto' to image filename'

jumbosushi:
## Problem
image

Editors tend to upload pictures with camera default name (ex. "zubair_XYZXYZYXZYXYZXYZ_ABCABCABC") which overflows the image picker preview div

image

Also the image-panel's have x-axis scroll bar by default when we don't need one. This takes away limited space in image-panel so should be removed. No padding right now in the form as well so things are kinda tight.

## After

image

Cleaner


Rowansdabomb (approved):
Nice!

PR #919, opened 2019-03-17
Add stripe overlay & use green border for img manager

jumbosushi:

image

I find myself always losing where the current focus is in the image manager UI.
 PR aims to make it visually easier to spot

image

Similarly, the insert btn is the primary action when inserting an img to an article, but the color isn't highlighted rn

image

- Changed border thickness from 2px to 4px
- Use green instead of red to communicate success state (or vibe?)
- Added stripe CSS overlay

image

- Make the insert btn be highlighted as primary action

## Before

image

## After

image

Feedback welcome


Rowansdabomb (approved):
LGTM

Issue #903, opened 2018-12-03
Add rss_feed link to podcast page


Rowansdabomb:
The rss feeds are not intuitive to find. A button that copies the link to the rss feed should be made available on the podcasts page. 

![untitled presentation](https://user-images.githubusercontent.com/19273530/49402110-d249a280-f6fd-11e8-8f12-9afb9ec7d877.png)

PR #922, opened 2019-03-23
Add rss feed copy to clipboard feature on podcasts

razvannesiu:

This PR solves the issue (#903 ) of not having an RSS feed on the podcasts page.

Created & used a CopyToClipboard Component, which uses the newly added Tooltip component. Tooltips can be activated on click or on hover. They can be positioned top, bottom, right, left.

### DIFF
--- a/dispatch/static/manager/src/js/components/CopyToClipboard.js
+++ b/dispatch/static/manager/src/js/components/CopyToClipboard.js
@@ -1,15 +1,21 @@
 import React from 'react'
+import Tooltip from '../components/Tooltip'
 
-class CopyToClipboard extends React.Component{
-  render(){
+require('../../styles/components/tooltip.scss')
+require('../../styles/components/copy_to_clipboard.scss')
+
+class CopyToClipboard extends React.Component {
+  render() {
     return (
-      <div>
+      <div className="c-copy-container">
         <textarea
-          ref={(node) => { this.txtInput=node }}
+          ref={(node) => { this.txtInput = node }}
           value={this.props.text} />
-        <button 
-          className="bp3-button bp3-icon-large bp3-icon-clipboard" 
-          onClick={() => { this.txtInput.select(); document.execCommand('copy')}} />
+        <Tooltip text={'Copied!'} position={'top'} activateOnClick>
+          <button
+            className="bp3-button bp3-icon-medium bp3-icon-clipboard"
+            onClick={() => { this.txtInput.select(); document.execCommand('copy') }} />
+        </Tooltip>
       </div>
     )
   }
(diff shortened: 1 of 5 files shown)


dispatch/static/manager/src/js/components/CopyToClipboard.js
```diff
@@ -0,0 +1,24 @@
...
+require('../../styles/components/copy_to_clipboard.scss')
+
+class CopyToClipboard extends React.Component {
+  render() {
+    return (
+      <div className="c-copy-container">
+        <textarea
+          ref={(node) => { this.txtInput = node }}
+          value={this.props.text} />
+        <Tooltip text={'Copied!'} position={'top'} activateOnClick>
```
Rowansdabomb: Explicitly setting `activateOnClick={true}` may be a little easier to read and won't rely on `undefined` for the expected behavior

dispatch/static/manager/src/js/components/Tooltip.js
```diff
@@ -0,0 +1,44 @@
...
+    
+    this.state = {
+      isTooltipDisplayed: false
+    }
+    
+    this.hideTooltip = this.hideTooltip.bind(this)
+    this.showTooltip = this.showTooltip.bind(this)
+  }
+
+  hideTooltip () {
```
Rowansdabomb, line 15: ES6 arrow functions give us a cleaner way to bind this to our react methods. 

`hideTooltip = () = > {` will automatically bind this to the context of the component and remove the need for binding in the constructor.

dispatch/static/manager/src/js/components/Tooltip.js
```diff
@@ -0,0 +1,44 @@
...
+
+class Tooltip extends React.Component{
+  constructor(props) {
+    super(props)
+    
+    this.state = {
+      isTooltipDisplayed: false
+    }
+    
+    this.hideTooltip = this.hideTooltip.bind(this)
```
Rowansdabomb, line 11: Can remove these binds if you do the step below.

dispatch/static/manager/src/js/components/Tooltip.js
```diff
@@ -0,0 +1,44 @@
...
+    this.showTooltip = this.showTooltip.bind(this)
+  }
+
+  hideTooltip () {
+    this.setState({isTooltipDisplayed: false})
+  }
+      
+  showTooltip () {
```
Rowansdabomb, line 19: Same thing here

dispatch/static/manager/src/js/components/Tooltip.js
Rowansdabomb: Also, I am in the process of updating the code base to utilize ES6 syntax, so it will be good practice to write new components with arrow functions

Rowansdabomb (changes requested):
This is awesome man, there are just a couple small changes I would make.

Rowansdabomb (approved):
LGTM

Issue #865, opened 2018-09-05
Add tooltip to display dispatch version under dispatch logo


Rowansdabomb:
Now that Phoenix is up and running, it will be useful to know which dispatch version is live at a glance

PR #923, opened 2019-03-25
Add version tooltip under dispatch logo

razvannesiu:

This PR solves the issue (#865 ) about displaying the dispatch version under the dispatch logo.

Added a tooltip conditionally, depending on whether we have the subscript available (the subscript in this context would be the dispatch version). The tooltip was added on the text "dispatch". Removed the underline effect of the tooltips.


Rowansdabomb (approved):
LGTM

PR #925, opened 2019-03-28
[Security] Update Django to 1.11.19

jumbosushi:
GitHub have been warning that we need to update django version for security reasons. 
Checked locally and seems to keep the app going the same way

https://github.com/ubyssey/dispatch/network/alert/setup.py/django/open


jumbosushi:
Travis failed as 1.11.19 doesn't exist on the pypi apparently https://pypi.org/simple/django/

jumbosushi:
lol

image

image

Rowansdabomb (approved):
LGTM

PR #926, opened 2019-04-15
New push notifications

Rowansdabomb:
Adds the ability to subscribe to push notifications from the ubyssey

#492

I mainly un-reverted the 492-push-notification branch. Additionally, I removed Django-Kronos as a dependency and I am hopeful we can achieve a cron job using gae's built-in cron.yaml

adds a schedule button to the article sidebar to let publishers choose when they want their article push notification to run.

comments out non push notification related code in service worker, as the code was using a depricated .waituntil() method which appears to break the service worker (stops it from becoming active).

adds a "secret" endpoint /beta/notifications from which subscriptions to ubyssey can be tested.


Rowansdabomb:
Closing due to inactivity

Issue #936, opened 2019-05-16
Video Page 


Rowansdabomb:
We need to add support for featured_videos in articles as well as add a video's page.

ghost:
this works.

PR #937, opened 2019-05-16
936 video

Rowansdabomb:
#936  Add Supports for Video article

Changed `Featured Image` tab to be `Featured Media` that allows user to choose one of the two

image

Shout out to Ron for laying out the ground work for this 
Shout out to Atsushi for his shout out to Ron


jumbosushi:
Thanks a lot for the help! What do you think about breaking apart the video migration file to separate migrations? I noticed that we have missed migrations for Articles prior to this PR, and it's being included in this video migration. 

Maybe to `0023_publishable_constraints_default` and `0024_video_meta_fields`

dispatch/modules/content/models.py
```diff
@@ -451,10 +451,33 @@ def get_absolute_url(self):
...
+    authors = ManyToManyField(Author, related_name='video_authors')
+    tags = ManyToManyField('Tag')
+
+    created_at = DateTimeField(auto_now_add=True)
+    updated_at = DateTimeField(auto_now=True)
+
+    AuthorModel = Author
+
+    # Overriding
+    def save(self, **kwargs):
```
jumbosushi: We aren't doing anything special in `save()` (yet) so we can remove this method

Rowansdabomb: Good catch! I forgot I removed the custom save operations

Rowansdabomb:
I like the idea of separating the migrations

jumbosushi:
We should add columns too for `/videos/`

image

Rowansdabomb:
I should have already added a preview, author, created_at, and updated_at
column. Here is a picture of how it looks on my machine.

[image: image.png]

*Rowan Skye*

PR #938, opened 2019-05-16
updated dispatch

ghost:

update to python 3

fixed dispatch admin


Rowansdabomb:
I am a bit unclear as to what this PR is doing... Where is `.idea/` directory? I do not see it in the develop branch, also why are you deleting it?

Rowansdabomb:
The title of this PR is ambiguous. Also, the description says this will update dispatch to python 3, but there are no changes to any python versioning that I can see.

dispatch/bin/dispatch-admin
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
Rowansdabomb, line 1: This is a comment is it not? does this actually cause any sort of change to the function of the code?

razvannesiu: It is not a comment, even though it starts with a #. It is called a [shebang line](https://stackoverflow.com/questions/2429511/why-do-people-write-the-usr-bin-env-python-shebang-on-the-first-line-of-a-pyt). When you execute dispatch-admin, that line determines what interpreter will run this code.

ghost:
We are going to close this PR and open a new one with all the files we have modified for the update. All the files from .idea had conflicts with develop branch so we removed them. Those are created by PyCharm.

PR #946, opened 2019-06-08
Upgrade Django to 2.2 LTS

razvannesiu:

This PR solves the issue of upgrading Django from 1.11 to 2.2, which is another LTS version.

I followed the docs for migration:
https://docs.djangoproject.com/en/2.2/releases/2.2/


ghost:
lgtm

PR #953, opened 2019-06-27
made full_name and slug unique and reqired

ghost:

made full_name and slug required and unique

changed blank value


razvannesiu (approved):
LGTM

PR #965, opened 2020-06-03
Settings revision

keeganland:

Difficulty with getting settings files and appropriate modern-day ("12 factor app") config standards, due to apps dispatch works with needing to outright import its settings file

Turned dispatch into a Django app so its files may be interacted with using DispatchConfig class in apps.py, rather than finding them relative to the settings file. This is a big API change, requiring us to release this as Dispatch 1.0.0.


davidkao0902:
Seems good to me.

PR #966, opened 2020-06-04
Upgrade djangorestframework

keeganland:

Django REST Framework out of date

Upgraded Django REST Framework according to official documentation


davidkao0902 (approved):
Seems good.

PR #967, opened 2020-06-04
Upgrade Django

keeganland:

Django out of date

Upgrade Django according to its official documentation


davidkao0902 (approved):
Seems good.

PR #969, opened 2020-06-11
Drop cap function

davidkao0902:
- Added drop cap function.


keeganland (approved):
You missed a version variable, but I got it

PR #975, opened 2020-07-02
Youtube url bug

keeganland:

Youtube URLs are parsed with assumptions that don't always hold

Use regex to get video ID rather than splitting string on '='


davidkao0902:
Looks good, but are the checks fine?

PR #976, opened 2020-07-18
Remove version variables

keeganland:

We've been manually versioning the static files. ManifestStaticFilesStorage can do this automatically. Therefore maintaining version variables will become redundant with future Ubyssey release and needs to be patched out of Dispatch

Removed version variables


davidkao0902 (approved):
lgtm

